Issue #8, opened 2024-02-29
Improve tests for peptides


pm-blanco:
Improvements:

- [x] Merge `GLU_HIS_peptide.py`,  `histatin5_peptide.py` and  `LYS_ASP_peptide.py`) should be merged into a single master script that runs all systems, similar to what is done for globular proteins in  `run_test_protein.py`.
- [x] Select a few interesting cases a built a numerical comparision compatible with the CI scheme

Issue migrated from our original remote repository on Gitlab, see here for further discusion on the topic [here](https://gitlab.com/blancoapa/pyMBE/-/issues/18)

Issue #13, opened 2024-03-01
Convert pyMBE into a package


jngrad:
The pyMBE module is currently not constructed as a *package*, thus limiting the ways in which the Python standard library can interact with it. For example, locating text file resources such as pKa datasets required the introduction of the class method `pymbe_library.get_resource()`, which takes a relative path as argument and joins it to the `__file__` path to create an absolute path. Making this method an internal method would force all pyMBE method to be only able to read datasets stored in the pyMBE source tree, which would force users to copy their custom datasets in the source tree directly.

To make pyMBE behave like a package, we could adapt its structure to look more like this:
```
./
 |---> pyMBE/
 |       |---> pyMBE.py
 |       |---> __init__.py
 |       |---> reference_parameters/
 |       |---> handy_scripts/
 |---> testsuite/
 |---> samples/
```

The `__init__.py` file would contain this code:

```py
from .pyMBE import *
```

The `pyMBE.py` file would be able to do relative imports of the form:

```py
from .handy_scripts.handy_functions import setup_electrostatic_interactions
```

Datasets can then be found using the [`importlib.resources`](https://docs.python.org/3/library/importlib.resources.html) mechanism, which replaces the deprecated [`pkg_resources`](https://setuptools.pypa.io/en/latest/pkg_resources.html) tool. Simulation scripts would work like this:

```py
import pyMBE
import importlib.resources
pmb = pyMBE.pymbe_library()
pmb.load_pka_set(filename=importlib.resources.files(pyMBE) / "reference_parameters" / "pka_sets" / "CRC1991.txt")
```

Notice how we no longer provide paths as strings containing the `'/'` character, but instead use the `/` operator which joins the path components using an OS-specific separator character (`'/'` on UNIX and `'\'` on Windows). You can try this out now with the following code:

```
$ python3 -c 'import pyMBE,importlib.resources as ir;print(ir.files(pyMBE)/"pka"/"CRC1991.txt")'
TypeError: <module 'pyMBE' from '/work/jgrad/pyMBE/pyMBE.py'> is not a package
$ echo "from .pyMBE import *" > __init__.py
$ cd .. # going up so Python can detect the "pyMBE" folder as a package and import its __init__.py
$ python3 -c 'import pyMBE,importlib.resources as ir;print(ir.files(pyMBE)/"pka"/"CRC1991.txt")'
/work/jgrad/pyMBE/pka/CRC1991.txt
[...]

PR #14, opened 2024-03-05
Add CI for peptides and clean-up folder structure

### LINKED ISSUE #3: Create a folder with all scripts to reproduce the data in our paper
Currently, there are some issues that we need to solve to make easy to reproduce the data from our paper:

- The scripts used to produce the data in our paper are scattered between `samples` and `tests` 
- The scripts for peptides are case-specific #18 
- We need a "main" script capable to run "tests/simulation_script_grxmc.py" to reproduce the data in `reference_data/data_landsgesell.csv` @davidbbeyer  

TODO:
- [x] Move all scripts necessary to reproduce the data to a new folder `/samples/Beyer2024`
- [x] Develop a script to reproduce the data in Beyer2024
- [x] Adapt the script to reproduce the data in Fig. 6 of Beyer2024
- [ ] Adapt the script to reproduce the data in Fig. 7 of Beyer2024
- [ ] Adapt the script to reproduce the data in Fig. 8 of Beyer2024
- [x] Add a Readme file explaining how to reproduce the data of our paper and with the citation.

pm-blanco:
Fixes #3 and #8. List of changes:
-  Merge all the case-specific scripts for peptides into one single script `samples/Beyer2024/peptide.py`
- Create a script to reproduce the data of our preprint `samples/Beyer2024/create_paper_data.py`
- Add a README file to explain how to reproduce the data from our preprint with these scripts
- Substitute the CI test by a CI test that checks that the above scripts still reproduce selected data points of all peptide cases.
- Clean-up the repository structure

@jngrad I would appreciate your opinion in the framework for the CI test and to reproduce the data of the paper. If you think is sensible, we could use it as template for the tests for globular proteins and polyelectrolytes in dialysis.


samples/Beyer2024/create_paper_data.py
```diff
@@ -0,0 +1,187 @@
...
+        sequence="E"*5+"H"*5
+    elif fig_label == "6c":
+        sequence="nDSHAKRHHGYKRKFHEKHHSHRGYc"
+    else:
+        raise RuntimeError()
+    pH_range = np.linspace(2, 12, num=21)
+    for pH in pH_range:
+        run_command=f"python3 {script_path} --sequence {sequence} --pH {pH} --mode {mode}"
+        print(run_command)
+        os.system(run_command)
```
jngrad, line 57: Here and elsewhere (including in files not affected by this PR): I would **strongly** recommend moving away from [`os.system()`](https://docs.python.org/3/library/os.html#os.system) (which directly exposes the shell), in favor of [`subprocess.check_output()`](https://docs.python.org/3/library/subprocess.html#subprocess.check_output) without `shell=True`. Rules for escaping characters with special meaning in paths (e.g. whitespace, slashes) are cumbersome and easy to forget; `subprocess` does all that for you. When a shell command fails, `os.system` doesn't care: it returns an OS-specific error code and Python executes the next line, possibly one of these `os.system(f"rm {path}/*")` which might not delete the files you expected if `path` depended on a failed shell command...

MWE:

```py
import os
import subprocess
os.system("ls important_file.bak") # is there a backup file?
print("rm important_file")         # oops!
subprocess.check_output(["ls", "important_file.bak"])
print("rm important_file")         # will never execute without a backup file
```
Output:
```
ls: cannot access 'important_file.bak': No such file or directory
rm important_file
ls: cannot access 'important_file.bak': No such file or directory
  File "/home/user/mwe.py", line 5, in <module>
    subprocess.check_output(["ls", "important_file.bak"])
  File "/usr/lib/python3.10/subprocess.py", line 421, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/usr/lib/python3.10/subprocess.py", line 526, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['ls', 'important_file.bak']' returned non-zero exit status 2.
```

Note that `subprocess.run()` would not throw an exception, since it returns an object to wrap the exception and provide more context, like the error code and the command line arguments, to help the user diagnose the issue. Also, some of the `rm path/*` commands can probably be replaced with a temporary folder context manager, which is automatically cleaned up when leaving the context.

jngrad (commented):
Overall looks good to me. The tree structure still looks a bit weird, but this can be dealt with later as part of #13. The subprocess calls refactoring can also be dealt with in a separate PR.

samples/Beyer2024/create_paper_data.py
pm-blanco: @jngrad if you do not mind that we deal with the sub-process refactoring in a separated PR, then I would proceed with this PR so we can have the new CI in main for continuing with the rest of the work.  I know that the tree structure is still not the best, I only wanted to have a clean structure for the article before we designed what you suggested in #13. If you are OK then we can merge this PR once you approve it.

jngrad (approved):

PR #22, opened 2024-03-12
21 make calculate hh donnan independent of espresso

davidbbeyer:
@pm-blanco I have adapted the function calculate_HH_Donnan so that it is completely independent of ESPResSo now. The input is now simply the a dictionary containing the concentrations of all macromolecular species in the system. Furthermore, I have removed the separate sample script on the unified G-RxMC method and simply added a command-line argument to the sample script, so one can select between the standard formulation and the one with unified ion types.


pm-blanco:
@davidbbeyer thank you for taking care of this issue, I will review it ASAP. Meanwhile, could you please provide a small for the CI to check this function?

davidbbeyer:
@pm-blanco Sure, I can do that. Maybe we can just test the implementation of the grand-reaction method (for an ideal gas) against this function?

pm-blanco:
@davidbbeyer I would do something even simpler than that. I would generate a set of reference data value for a couple of cases, save it in the data folder of the testsuite and I would do a test test runs that function and checks that it still produces the same data. In this way it can be a very fast test for the CI. Please, if you follow this approach, save the data in the format than the other files there for consistency.

pm-blanco:
@davidbbeyer could you please also change the argument `object_name` in `calculate_HH` for `molecule_name`?

pm-blanco (approved):
The current implementation solves the issue and it provides an extra improvement in our CI testing to check the theoretical calculation of the ideal ionization for both single phase and two phase system. To me it is now good enough to be merged, thank you @davidbbeyer

PR #23, opened 2024-03-12
Build and deploy documentation with CI/CD pipeline

jngrad:
The documentation is now built as part of the CI and uploaded to an artifact. Broken docstrings will fail CI. When a PR is merged, the artifacts are downloaded and deployed to the GitHub Pages at https://pm-blanco.github.io/pyMBE.


jngrad:
Note to reviewers:
* deployment only happens to Pages of the main repository, not on forks
* it is not possible to accidentally deploy to Pages from a branch other than `main` thanks to deployment protection rules: `Branch "other" is not allowed to deploy to github-pages due to environment protection rules.`
* it is possible to bypass the deployment protection rules by manually eliding the `environment` field and pushing the workflow to any branch of the main repository (only collaborators can do that)
* however, malicious actors cannot use that trick when working from a fork, because they will have to open a PR, and the deployment workflow will fail with `Ensure GITHUB_TOKEN has permission "id-token: write"`

jngrad:
@pm-blanco To enable [GitHub Pages](https://docs.github.com/en/pages/getting-started-with-github-pages/about-github-pages), go to this repository "Settings", under "Code and automation" click "Pages", under "Build and deployment" select "GitHub Actions". Then you could can merge this PR; a testsuite workflow will start on the main branch, and once it completes successfully, it will send a signal to start the deploy workflow, which will fetch the status of the testsuite and download its HTML files, then upload them to the Pages.

We can then add https://pm-blanco.github.io/pyMBE to the manuscript. Note that the link automatically redirects to a subpage, which is the expected behavior from pdoc (it generates an index.html landing page), and I would recommend not writing the subpage URL in the manuscript, because this link would break if we re-organize the pyMBE tree structure. Currently https://pm-blanco.github.io returns a 404 because you do not have a personal page, but as far as I can tell, this has no consequence on the user guide.

pm-blanco (approved):
Sounds good to me, I did the changes in the settings. Let's merge this PR and see how our new docs look like!

PR #24, opened 2024-03-13
added value error if pka not provided for acidic or basic particles

### LINKED ISSUE #6: Users can define particles with acidic or basic properties without a pKa
This feature can ultimately lead to a crash of the simulation if the user does not define the pKa value of the particle later (for example loading a pka set). 

In `set_particle_acidity`, when the user provides an acidity "acidic" or "basic" pyMBE should check that it has also provided a pKa. If not, it should raise a ValueError an inform the user that it should provide it.

mariusaarsten:
Fixes #6


mariusaarsten:
@pm-blanco I have added a very simple unit test now. I see there are some conflicting files with the parameters, I don't think I have touched them, but should I just change them to what they are in the main fork?

pm-blanco:
@mariusaarsten yes, you will need to address the merge conflict:
- for `Blanco2020.txt` and `Lunkad2021.txt` just keep the version in the main fork
- for `Makefile` you need to merge it so it also includes the new test you added.

pm-blanco:
@mariusaarsten I think the CI is now breaking actually due to a broken setup in the peptide files, for example in `Lunkad2020.txt`:
```json
{"object_type":"particle", "name": "D", "acidity": "acidic", "sigma": {"value":0.35, "units":"nm"}, "epsilon":{"value":1, "units":"reduced_energy"}}
```
an acidity is given but no pKa value is provided because it is later loaded from the pKa set, which now raises a ValueError with your new implementation. I think the solution is simple: just remove the acidity from all particles in the peptide files.

pm-blanco:
@mariusaarsten I think the problem is that you need to change `diameter` to `sigma` in `Blanco2020.txt`

pm-blanco (approved):
It looks good to me now. Thank you for your work @mariusaarsten

PR #27, opened 2024-03-19
Refactor lj parameters

### LINKED ISSUE #19: Change logic of the LJ parameters
`diameter is not a fortunate choice because it actually defines two parameters provided to Espresso, and the connection between the two is hard-coded in pymbe. This may cause trouble in future, so we should reconsider changing it before publishing the paper. I think that the user should provide sigma and offset separately. Request by @kosovan .

pm-blanco:
Fixes #19 

List of changes:

- `define_particle` now takes  `sigma`,  `cutoff` and `offset` as optional arguments.
- `sigma` defaults to None. No LJ interactions are set up for particles with `sigma` or `epsilon` of None, maintaining the same behavior as before. The user gets a warning message informing that no LJ have been set up for those type of particles.
- If the user sets `sigma=0`, no LJ interactions are set for those particles corresponding to ideally behaving particles, maintaining the same behavior as before.
- `cutoff` defaults to `2**(1./6) reduced_length`, maintaining the same behavior as before.
- `offset`  defaults to 0. This introduces a new behavior as the user now needs to set-up on its own the offset instead of hard-coding that `sigma = 1 reduced_length` 
- A new CI test has been introduced `testsuite/lj_tests.py` with several unit tests for  `define_particle`  and `setup_lj_interactions`


samples/Beyer2024/peptide.py
pinedaps, line 1: Can we enhance the homogeneity of the code by substituting the properties referring to Na and Cl with 'cation' and 'anion' respectively?

pyMBE.py
```diff
@@ -457,6 +457,23 @@ def center_molecule_in_simulation_box(self, molecule_id, espresso_system):
...
+        """
+
+        try:
+            correct_dimensionality=variable.check(f"[{expected_dimensionality}]")
+        except:
+            raise ValueError(f"The variable {variable} should be a `pint.Quantity` object, instead the variable is {type(variable)}")
+        
+        if not correct_dimensionality:
+            raise ValueError(f"The variable {variable} should have a dimensionality of {expected_dimensionality}, instead the variable has a dimensionality of {variable.dimensionality}")
+
```
pinedaps, line 476: One needs to be careful with this function, so far it works only for single dimensionality ('time', 'length' ...) but it fails when passing complex dimensionality ('length/time'), and the provided a confusing error message:

```python
pmb.check_dimensionality(1*pmb.units.nm, 'time/length')
raise ValueError(f"The variable {variable} should be a `pint.Quantity` object, instead the variable is {type(variable)}")
ValueError: The variable 1 nanometer should be a `pint.Quantity` object, instead the variable is <class 'pint.Quantity'>
```
Additionally, I misunderstood the term 'dimensionality' at my first try and I set my expected_dimensionality to 'nanometer', It raised the same confusing message error.

```python
pmb.check_dimensionality(1*pmb.units.nm, 'nanometer')
raise ValueError(f"The variable {variable} should be a `pint.Quantity` object, instead the variable is {type(variable)}")
ValueError: The variable 1 nanometer should be a `pint.Quantity` object, instead the variable is <class 'pint.Quantity'>
```
It seems that this function is primarily utilized internally. However, it may be beneficial to enhance the documentation by providing an example, such as 'length', to illustrate what dimensionality is. It should be clarified that the function operates solely on single dimensionality.

pyMBE.py
```diff
@@ -1193,31 +1211,62 @@ def define_molecule(self, name, residue_list):
         self.df.at [index,('residue_list','')] = residue_list
         return
 
-    def define_particle(self, name, q=0, diameter=None, acidity='inert', epsilon=None, pka=None):
+    def define_particle(self, name, q=0, acidity='inert', pka=None, sigma=None, epsilon=None, cutoff=None, offset=None):
```
pinedaps, line 1214: In the docs says: `cutoff` defaults to `2**(1./6.) reduced_length` but in the function is None.

tutorials/pyMBE_tutorial.ipynb
```diff
@@ -1137,11 +2442,24 @@
...
+      "\u001b[0;31mFileNotFoundError\u001b[0m: [Errno 2] No such file or directory: '/home/pablo/Escritorio/pyMBE/reference_parameters/interaction_parameters/Lunkad2021.txt'"
+     ]
+    }
+   ],
+   "source": [
+    "pmb.load_interaction_parameters (filename = pmb.get_resource('parameters/peptides/Lunkad2021.txt'))\n",
```
jngrad, line 2462: Consider clearing the notebook cell outputs. If we need notebooks with cell outputs, we can try adding an extra step in the CI/CD pipeline to execute the notebooks and push them as static HTML files to the GitHub Pages.

pyMBE.py
```diff
@@ -2648,34 +2699,33 @@ def setup_df (self):
...
+        Sets up the Lennard-Jones (LJ) potential between all pairs of particle types with values for `sigma`, `offset`, and `epsilon` stored in `pymbe.df`.
 
         Args:
             espresso_system(`obj`): Instance of a system object from the espressomd library.
-            cutoff(`float`, optional): cut-off length of the LJ potential. Defaults to None.
-            shift (`string`, optional): If set to `auto` shifts the potential to be continous at `cutoff`. Defaults to `auto`.
-            combining_rule (`string`, optional): combining rule used to calculate `sigma` and `epsilon` for the potential betwen a pair of particles. Defaults to 'Lorentz-Berthelot'.
-            warning (`bool`): switch to activate/deactivate warning messages. Defaults to True.
+            shift_potential(`bool`, optional): If True, a shift will be automatically computed such that the potential is continuous at the cutoff radius. Otherwise, no shift will be applied. Defaults to True.
+            combining_rule(`string`, optional): combining rule used to calculate `sigma` and `epsilon` for the potential betwen a pair of particles. Defaults to 'Lorentz-Berthelot'.
```
pinedaps, line 2709: typo= between

testsuite/lj_tests.py
```diff
@@ -0,0 +1,171 @@
...
+    for parameter_key in ["sigma","offset","cutoff"]:
+        np.testing.assert_equal(actual=setup_lj_parameters[parameter_key], 
+                                desired=ref_lj_parameters[parameter_key].to("reduced_length").magnitude, 
+                                verbose=True)
+    np.testing.assert_equal(actual=setup_lj_parameters["epsilon"], 
+                                desired=ref_lj_parameters["epsilon"].to("reduced_energy").magnitude, 
+                                verbose=True)
+print(f"*** Unit test passed ***")
+
+print(f"*** Unit test: test that setup_lj_interactions does not setup any LJ interactions for particles with sigma = 0 ***")
```
pinedaps, line 160: typo = does not set up

pinedaps (approved):
I believe the implementation was carried out correctly. I left a few minor comments, but they are not critical in any way. The unit test appears to be robust enough.

pyMBE.py
pm-blanco: @pinedaps thank you for pointing out this issue:
- I improved the function to work with variables with complex dimensionalities and I improved the documentation of the function to explain the expected format for `expected_dimensionality`.
-  I removed the ValueError and instead I let Pint throw its own internal error since it is actually a more clear message error.

tutorials/pyMBE_tutorial.ipynb
pm-blanco: @jngrad good catch! this was unintended, I will clean the cell outputs before merging.

pinedaps (approved):
Perfect!

pinedaps (approved):
Perfect!

PR #33, opened 2024-04-14
Add test of G-RxMC (ideal).

davidbbeyer:
Solves #32 

@pm-blanco I have added a test that checks if the result of the G-RxMC implementation (both standard and unified) is consistent with the Henderson-Hasselbalch + Don. If you agree with the overall conception, I would add the same for the cpH-ensemble before we merge it.

### DIFF
--- a/samples/peptide_mixture_grxmc_ideal.py
+++ b/samples/peptide_mixture_grxmc_ideal.py
@@ -8,7 +8,6 @@ import numpy as np
 import matplotlib.pyplot as plt
 import pandas as pd
 import argparse
-from tqdm import tqdm
 from espressomd.io.writer import vtf
 from espressomd import interactions
 from espressomd import electrostatics
@@ -19,7 +18,7 @@ pmb = pyMBE.pymbe_library()
 
 # Command line arguments
 
-valid_modes=["standard", "unified"]
+valid_modes=["standard", "unified", "standard-test", "unified-test"]
 parser = argparse.ArgumentParser(description='Script that runs a simulation of an ideal peptide mixture in the grand-reaction ensemble using pyMBE and ESPResSo.')
 parser.add_argument('--mode',
                     type=str,
@@ -106,7 +105,7 @@ pmb.define_peptide (name=peptide2, sequence=sequence2, model=model)
 # Solution parameters
 c_salt=5e-3 * pmb.units.mol/ pmb.units.L
 
-if args.mode == 'standard':
+if args.mode == 'standard' or args.mode == 'standard-test':
     proton_name = 'Hplus'
     hydroxide_name = 'OHminus'
     sodium_name = 'Na'
@@ -117,7 +116,7 @@ if args.mode == 'standard':
     pmb.define_particle(name=sodium_name, q=1, sigma=0.35*pmb.units.nm, epsilon=1*pmb.units('reduced_energy'))
     pmb.define_particle(name=chloride_name,  q=-1, sigma=0.35*pmb.units.nm,  epsilon=1*pmb.units('reduced_energy'))
 
-elif args.mode == 'unified':
+elif args.mode == 'unified' or args.mode == 'unified-test':
     cation_name = 'Na'
     anion_name = 'Cl'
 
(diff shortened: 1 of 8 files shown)


pm-blanco:
@davidbbeyer thank you for your initiative! I agree with the overall concept, but I would change a bit the implementation, comments will follow

samples/peptide_mixture_grxmc_ideal.py
```diff
@@ -264,27 +263,41 @@
...
+    # Calculate the ideal titration curve of the peptide with Henderson-Hasselbach equation
+    HH_charge_dict = pmb.calculate_HH_Donnan(c_macro={peptide1: pep1_concentration, peptide2: pep2_concentration}, c_salt=c_salt, pH_list=pH_range)
+    Z_HH_Donnan = HH_charge_dict["charges_dict"]
+    xi = HH_charge_dict["partition_coefficients"]
+    
+    # Check if charges agree
+    np.testing.assert_allclose(np.asarray(Z_pH)/N_peptide1_chains, np.asarray(Z_HH_Donnan[peptide1])+np.asarray(Z_HH_Donnan[peptide2]), rtol=0.01, atol=0.05)
+
+    # Check if partition coefficients agree
+    np.testing.assert_allclose(xi_plus, np.asarray(xi), rtol=0.1, atol=0.1)
```
pm-blanco, line 303: I would move this lines to `testsuite/grxmc_ideal_tests.py` . In general, I would avoid test-specific in the sample scripts as much as possible because they are meant to be for the users to learn and I think that it will only confuse them.

samples/peptide_mixture_grxmc_ideal.py
```diff
@@ -264,27 +263,41 @@
...
-fig, ax = plt.subplots(figsize=(10, 7))
-plt.errorbar(pH_range, xi_plus, yerr=err_xi_plus, fmt = 'o', capsize=3, label='Simulation')
-ax.plot(pH_range_HH, np.asarray(xi), "-k", label='HH+Donnan')
-plt.legend()
-plt.xlabel('pH')
-plt.ylabel(r'partition coefficient $\xi_+$')
-plt.show()
-plt.close()
+
+if args.mode == 'standard' or args.mode == 'unified':
```
pm-blanco, line 267: In general I would try to avoid to have too many "test-specific" features in the codes of the samples, because they are meant to be examples for our users. 
I would try to refactor this a bit as follows:
- drop the test specific modes
- instead add a boolean flag `plot` to control if the script does the plotting or not, so you can deactivate that part when running the test (all the rest is actually not test specific)
- save the data to some files that can be then be read by the test and check that both produce the same result

samples/peptide_mixture_grxmc_ideal.py
```diff
@@ -19,7 +18,7 @@
 
 # Command line arguments
 
-valid_modes=["standard", "unified"]
+valid_modes=["standard", "unified", "standard-test", "unified-test"]
```
pm-blanco, line 21: please add a sanity check that checks that the user provide a valid mode, for example see 
`samples/Beyer2024/peptide.py

samples/peptide_mixture_grxmc_ideal.py
```diff
@@ -106,7 +105,7 @@
 # Solution parameters
 c_salt=5e-3 * pmb.units.mol/ pmb.units.L
 
-if args.mode == 'standard':
+if args.mode == 'standard' or args.mode == 'standard-test':
```
pm-blanco, line 108: and lines 119,142,147,166, 168. I think that you can drop the  test-related checks because it only matters if you are running the test for the last lines of the script

pm-blanco (changes requested):

davidbbeyer:
@pm-blanco I have implemented your suggestions

pm-blanco:
@davidbbeyer I think the current implementation looks very good, please do a similar test for the cpH and we can proceed to merge this PR.

davidbbeyer:
@pm-blanco I have added a test on cpH

pm-blanco:
@davidbbeyer the CI now breaks at the peptide tests, but they run locally for me. Maybe try increasing a bit the number of samples for the peptide tests to avoid ramdom crashes of the CI?

pm-blanco (approved):
It looks good to me now, thank you @davidbbeyer

PR #34, opened 2024-04-16
Fix initial pos mol

### LINKED ISSUE #35: bug:  create_molecule(positions) only creates the fist molecule in the correct position
the optional argument `position` should take a list of list with the positions where the first particle of each molecule should be created. Currently, it only has the expected behavior for the first molecule while the rest are created at random positions

mariusaarsten:
Fixes #35


pyMBE.py
```diff
@@ -722,101 +722,110 @@ def create_counterions(self, object_name, cation_name, anion_name, espresso_syst
...
+            used_molecules_id = self.df.molecule_id.dropna().drop_duplicates().tolist()
+            pos_index = 0 
+            for molecule_index in molecule_index_list:        
+                molecule_id = self.assign_molecule_id (name=name,pmb_type='molecule',used_molecules_id=used_molecules_id,molecule_index=molecule_index)
+                for residue in residue_list:
+                    if first_residue:
+                        if list_of_first_residue_positions == None:
+                            residue_position = None
+                        else:
+                            if len(list_of_first_residue_positions) != number_of_molecules:
```
pm-blanco, line 763: This sanity test should be moved to the begging of the function (there is no need to check for each loop iteration)

testsuite/create_molecule_position.py
```diff
@@ -0,0 +1,84 @@
+import numpy as np
```
pm-blanco, line 1: note that you have added two tests that do the same (or very similar) `testsuite/create_molecule_position.py` and `testsuite/create_molecule_position_test.py` please either remove one of them or if they do something different merge them into one

pm-blanco (changes requested):

pm-blanco:
@mariusaarsten note that you need to change the name of the test in the `Makefile`, otherwise the CI crashes simply because it does not find the right file

mariusaarsten:
@pm-blanco Done! I think the pull request should be good now

pm-blanco (approved):
@mariusaarsten it looks good to me now, thank you for your work!

PR #39, opened 2024-04-21
Improve proteins script 

paobtorres:
I created the PR but there are a few points to discuss before the merge:

- I adapted the script to add the radius of each beach (If im not mistaked sigma = 2*radius right?).  However the results have a wider difference compared to Torres2017 and Torres2022 in some of the pH values. (I checked running in  `test` and `short-run` mode). Also I get the warning the I'm overwritting the sigma of a previous defined aminoacid I checked a few but It seems to be ok.
- In `globular_protein.py` I added to the time_series output the charge of each aminoacid but I'm not sure how is the correct way. Currently:
` for label in net_charge_amino_save.keys ():
        charge_amino = sum(net_charge_amino_save[label])
        time_series[label].append(charge_amino)` 
I'm not sure if it should be sum or mean, I tried both but non of them seem to be correct. 
-While doing this I notice that In `charge_dict ['residues']` there were several `CA` that had 1 positive charge. I don't know why and if it only happens in the proteins. 
- `create_paper_data.py` its working for the protein If you would like to check 
- `testsuite\globular_protein_tests.py` Its also working but I would like to fix the bugs that I mentioned before uploading the test data


paobtorres:
@pm-blanco I finished with all the changes related to the radius. I run a simulation on test mode an the plots looks much much better. Also, I already merge this branch with the lastest changes from Seba PR. 
I think the only thing missing is the AA net charge that I didn't have the time to check (maybe tomorrow morning if you don't do it first).

pm-blanco (approved):
I think that the current version looks good, thank you for your work @paobtorres

PR #41, opened 2024-04-22
Bonds

### LINKED ISSUE #20: Refactor how bonds are defined in pyMBE
Currently, `define_bond` follows a different logic than the rest of functions to define objects on pyMBE. `define_bond`  should:

- [x] get `name`,  `bond_parameters`, `bond_type` and  `bonded_particle_names` as keyword arguments.
- [x] Define the curresponding espresso object internally
- [ ] Then, pmb.add_bonds_to_es should create bond objects in espresso

Request by @kosovan.

pinedaps:
Fixes #20 

List of changes:

The arguments in the function 'create_bond_in_espresso' reduced to two: the string, 'bond_type' and a dictionary, 'bond_parameters' which should contain the parameters for the corresponding 'bond_type', either harmonic or FENE.

Because of the reduction of arguments in the function 'create_bond_in_espresso', the argument of the functions 'define_bond' and define_default_bond' were reduced in the same way. Additionally, now the function 'define_bond' should be called only once by providing a list of the pair of particle types to be bonded in the argument 'particle_pairs'.

All the tests were updated as well as the tutorials.

Finally, the script 'bond_tests.py' is the unit test for checking that all functions involved in the definition of bonds are working properly.


pm-blanco (approved):
@pinedaps I did some clean-up in the test, adapted the load_interactions functions to take the same arguments as define_bond when reading from file and I fix a bug when defining FENE bonds. It looks good to me know, if you agree with the changes we can proceed and merge this PR. Thank you for your work!

pinedaps (commented):
@pm-blanco I just checked your changes and now it looks perfect!

PR #46, opened 2024-04-27
Fix center molecule

paobtorres:
Adapted center_molecule for non cubic boxs


pm-blanco (changes requested):
@paobtorres can you please also add a CI test to check that this feature works properly? it could be included within the test `testsuite/create_molecule_position_test.py` and it should check that `center_molecule_in_simulation_box` creates a molecule in the center of the simulation box for both a cubic a non-cubic simulation box

Issue #37, opened 2024-04-19
Add a function to set up GCMC simulations


davidbbeyer:
We should add a function to set up simulations of systems grand-canonically coupled to a reservoir of salt.

PR #47, opened 2024-04-27
Add an implementation of GCMC.

davidbbeyer:
Solves #37 

@pm-blanco I have added a function `setup_gcmc` that allows for setting up simulations where the system is coupled grand-canonically to a reservoir at a given salt concentration. The design of the function is similar to the setup of the grand-reaction method; optionally one can provide again a function that calculates the activity coefficient in order to correctly set up the method in the case of an interacting system. 

A minimal example is provided in `samples/salt_solution_gcmc.py`; this script is also used in two tests I added to the testsuite, one of them for an ideal system and the other one for an interacting system. Both of these tests check if the GCMC implementation leads to the expected (input) concentration of salt.

### DIFF
--- a/testsuite/gcmc_tests.py
+++ b/testsuite/gcmc_tests.py
@@ -8,16 +8,30 @@ import numpy as np
 import pandas as pd
 import argparse
 
+# Template of the test
+
+def gcmc_test(mode):
+    if mode == "ideal":
+        print(f"*** Running test for GCMC of salt solution (ideal). ***")
+    elif mode == "interacting":
+        print(f"*** Running test for GCMC of salt solution (interacting). ***")
+    with tempfile.TemporaryDirectory() as time_series_path:
+        for c_salt_res in salt_concentrations:
+            print(f"c_salt_res = {c_salt_res}")
+            run_command=["python3", script_path, "--c_salt_res", str(c_salt_res), "--output", time_series_path, "--mode", mode, "--no_verbose"]
+            print(subprocess.list2cmdline(run_command))
+            subprocess.check_output(run_command)
+        # Analyze all time series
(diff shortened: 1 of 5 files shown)


testsuite/gcmc_tests.py
```diff
@@ -0,0 +1,45 @@
...
+elif args.mode == "interacting":
+    print(f"*** Running test for GCMC of salt solution (interacting). ***")
+with tempfile.TemporaryDirectory() as time_series_path:
+    for c_salt_res in salt_concentrations:
+        print(f"c_salt_res = {c_salt_res}")
+        run_command=["python3", script_path, "--c_salt_res", str(c_salt_res), "--mode", "ideal", "--output", time_series_path, "--mode", args.mode, "--no_verbose"]
+        print(subprocess.list2cmdline(run_command))
+        subprocess.check_output(run_command)
+    # Analyze all time series
+    data=analysis.analyze_time_series(path_to_datafolder=time_series_path)
```
pm-blanco, line 39: @davidbbeyer I would prefer to change the logic of the test a bit to avoid having to call the script twice:
1) Encapsulate this lines into a function (for example `test_gcmc`)
2) remove the argparse input
3) simply call the function twice in the script for the interacting and the ideal case.
The reason is that then in the Makefile each line correspond to call to a different testing block, preventing it to grow too much in the future when we add more tests.

Otherwise this PR looks very good to me.

pm-blanco (changes requested):

testsuite/gcmc_tests.py
davidbbeyer: @pm-blanco I have adjusted the test accordingly

pm-blanco (approved):
looks good to me, thank you @davidbbeyer

Issue #45, opened 2024-04-27
Make the ionic product of water an attribute of pyMBE


davidbbeyer:
Currently, the ionic product of water KW=1e-14 (used in the setup of the G-RxMC method) is fixed and the value cannot be changed by the user. This can lead to inconsistent results when running simulations at temperatures different than room temperature. Hence, KW should be changeable by the user like the temperature.

PR #48, opened 2024-04-29
Make the ionic product an attribute of pyMBE

davidbbeyer:
Solves #45 @pm-blanco

### DIFF
--- a/pyMBE.py
+++ b/pyMBE.py
@@ -9,6 +9,7 @@ class pymbe_library():
         e(`obj`): Elemental charge using the `pmb.units` UnitRegistry.
         df(`obj`): PandasDataframe used to bookkeep all the information stored in pyMBE. Typically refered as `pmb.df`. 
         kT(`obj`): Thermal energy using the `pmb.units` UnitRegistry. It is used as the unit of reduced energy.
+        Kw(`obj`): Ionic product of water using the `pmb.units` UnitRegistry. Used in the setup of the G-RxMC method.
     """
     import pint
     import re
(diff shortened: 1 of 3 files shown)


pm-blanco:
@davidbbeyer I think that it would be better to name the varible consistently `Kw` rather than having different variable names in 
```python
def __init__(self, temperature=None, unit_length=None, unit_charge=None, ionic_product=None)
```
and in `pmb.Kw`.  Also the name `ionic_product` sounds a bit vague to me, as it could refer to some other ionic product than that or water (for example for a precipitation reaction). 

Could you clean the history of your repository? It seems that you have some old commits from the GCMC implementation that are squashed in the principal repo.

kosovan:
I strongly support naming the variables as close as possible to the symbols used in Equations.

kosovan:
I also agree that `ionic_product` is much more ambiguous than Kw.

davidbbeyer:
@pm-blanco I have adjusted the nomenclature accordingly. It is probably easiest if you just squash it before merging. I will properly sync my fork after the merge then.

pm-blanco:
@davidbbeyer the problem with squashing this commits is that then we will have repeated commits in two squashed PR (this one and the previous one) which will mess the history of our repository. I know is a bit tedious, but I think that it would be better if you can solve it before merging this to the main repo.

davidbbeyer:
@pm-blanco Now it should be cleaned up properly :)

pm-blanco (approved):
it looks good to me

Issue #44, opened 2024-04-27
SEED should be a global variable


pm-blanco:
The seed for the RNG should be a mandatory argument of  `__init__`   so that one provides it when creating an instance of pyMBE:
```python
pmb = pyMBE_library(SEED=42)
```

and it is internally stored as an attribute of `pmb`

```python
__init__(self, SEED, temperature=None, unit_length=None, unit_charge=None):
        self.SEED=SEED
```

then all methods of pyMBE should access to this value instead of having multiple SEEDs values. This implementation would make debugging of scripts using pyMBE much easier and would also also make our CI tests more consistent.

Thank you @kosovan for reporting this issue.

pm-blanco:
@davidbbeyer maybe you can take a look at this?

davidbbeyer:
@pm-blanco Yes, I was actually planning to make this my next contribution :)

Issue #43, opened 2024-04-26
pmb.center_molecule_in_simulation_box fails for non-cubic boxes


kosovan:
Current implementation

box_center = [espresso_system.box_l[0]/2.0]*3

should be replace with

box_center = [
                espresso_system.box_l[0]/2.0,
                espresso_system.box_l[1]/2.0, 
                espresso_system.box_l[2]/2.0
                ]

pm-blanco:
@paobtorres could you please take care of this?

PR #50, opened 2024-05-03
Make the seed a global variable

davidbbeyer:
Solves #44, #43 

@pm-blanco I have made the `SEED` a mandatory attribute of the `pmb` class, which is now used to set up the various reaction methods. Furthermore, I have added an attribute `rng` that is initialized using the provided `SEED` and used in `generate_random_points_in_a_sphere`. This way it is avoided that the `rng`, which was before initialized within `generate_random_points_in_a_sphere`, is each time re-initialized with the same seed. 

All samples, tests, etc. have been updated accordingly.

Moreover, I have also fixed the bug discovered by @kosovan regarding non-cubic simulation boxes (#43)

### DIFF
--- a/pyMBE.py
+++ b/pyMBE.py
@@ -24,8 +24,9 @@ class pymbe_library():
     df=None
     kT=None
     Kw=None
+    SEED=None
 
-    def __init__(self, temperature=None, unit_length=None, unit_charge=None, Kw=None):
+    def __init__(self, SEED, temperature=None, unit_length=None, unit_charge=None, Kw=None):
         """
         Initializes the pymbe_library by setting up the reduced unit system with `temperature` and `reduced_length` 
         and sets up  the `pmb.df` for bookkeeping.
@@ -42,6 +43,8 @@ class pymbe_library():
             - If no `unit_charge` is given, a value of 1 elementary charge is assumed by default. 
             - If no `Kw` is given, a value of 10^(-14) * mol^2 / l^2 is assumed by default. 
         """
+        # Seed for RNG
+        self.SEED=SEED
         # Default definitions of reduced units
         if temperature is None:
             temperature= 298.15 * self.units.K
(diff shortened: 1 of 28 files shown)


pm-blanco:
thanks @davidbbeyer for taking care of this, we should make a couple of CI tests for this:
- one that tests that `center_molecule_in_simulation_box` places the molecule in the center of the simulation box, for both cubic and non cubic simulation boxes. Probably could be part of `testsuite/create_molecule_position_test.py`
- We should have a CI test that checks that pyMBE produces deterministic results for a given `SEED`.  I am not completely sure what is the best way we could check this. A simple test could be to create a somewhat complex molecule twice into espresso with two different instances of pyMBE with the same `SEED` and check that one gets the same structure

davidbbeyer:
@pm-blanco I have implemented your suggestions:

- `testsuite/create_molecule_position_test.py` now contains two tests that check if the function `center_molecule_in_simulation_box` works correctly. Note that I had to slightly modify the function `center_molecule_in_simulation_box`, since it was performing a check that was not strictly necessary and in some case might cause that the molecule is _not_ centered in the box (depending on the order that objects with a given molecule id are added to the dataframe). I also adjusted the function `create_molecule` such that the returned dictionary now actually matches the description in the docstring.

- I added a test `testsuite/seed_test.py` that checks if a molecule is created at exactly the same position if the same global seed is provided to pyMBE.

pyMBE.py
```diff
@@ -450,15 +455,14 @@ def center_molecule_in_simulation_box(self, molecule_id, espresso_system):
             molecule_id(`int`): Id of the molecule to be centered.
             espresso_system(`obj`): Instance of a system object from the espressomd library.
         """
-        center_of_mass = self.calculate_center_of_mass_of_molecule ( molecule_id=molecule_id,espresso_system=espresso_system)
-        box_center = [espresso_system.box_l[0]/2.0]*3
-        pmb_type = self.df.loc[self.df['molecule_id']==molecule_id].pmb_type.values[0]
-        pmb_objects = ['protein','molecule','peptide']
-        if pmb_type in pmb_objects:
```
pm-blanco, line 457: You are right that this check is not really necessary, However, what is missing is a sanity check to ensure that the input molecule_id is defined in the pmb.df, otherwise the user will get a somewhat confusing pandas error

pm-blanco (changes requested):
Apart from this minor things, it looks very good to me. Thank you for taking care of the bug in the return info in `create_molecule`

testsuite/seed_test.py
```diff
@@ -0,0 +1,61 @@
+import numpy as np 
+import espressomd
+import warnings
+from espressomd import interactions
+import pyMBE
+
+espresso_system = espressomd.System(box_l = [100]*3)
+
+def run_peptide_simulation(SEED):
```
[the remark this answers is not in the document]
davidbbeyer: done

pyMBE.py
davidbbeyer: I have added such a check.

pm-blanco (approved):
it looks good to me

PR #55, opened 2024-05-07
Add Pylint

jngrad:
Description of changes:
- fix bugs in `assign_molecule_id()` and sample scripts
- resolve Pylint issues with high severity
- add Pylint checks to the CI pipeline
- make code more future-proof
- address MPI and `pdoc` warnings in CI


testsuite/bond_tests.py
```diff
@@ -1,36 +1,36 @@
...
 pmb = pyMBE.pymbe_library(SEED=42)
 
 def check_bond_setup(bond_object, input_parameters, bond_type):
-        """
-        Checks that pyMBE sets up a bond object correctly.
-
-        Args:
-                bond_object(`espressomd.interactions`): instance of a espresso bond object.
-                input_parameters(`dict`): dictionary with the parameters for the bond potential.
-                bond_type(`str`): label identifying the bond type
```
jngrad, line 15: When visualizing the diff on GitHub, be sure to click on "Hide whitespaces" under the gear icon to remove clutter. In the terminal, use `git show --color-words HEAD^` to remove whitespace (also works with `git diff`). In some files, you might see extra `^M` at the end of lines: some files on the main branch were saved with Windows line endings (`CRLF`) instead of Unix line endings (`LF`). I chose not to change line endings to Unix in this PR, otherwise the diff would have become unreadable. Maybe someone else can volunteer to go through all scripts and change line endings in a separate PR, using [`dos2unix`](https://packages.ubuntu.com/noble/dos2unix) or text editors like `gedit`. Right now copy-pasting a code sample from a Windows file into a Unix file can lead to extra newline characters appearing, depending on the text processor you are using. In fact, some of the scripts in pyMBE have double newlines in a few code blocks, probably because of copy-paste issues.

pyMBE.py
```diff
@@ -186,8 +186,8 @@ def assign_molecule_id(self, name, molecule_index, pmb_type, used_molecules_id):
         else:
             # check if a residue is part of another molecule
             check_residue_name = self.df[self.df['residue_list'].astype(str).str.contains(name)]
-            pmb_type = self.df.loc[self.df['name']==name].pmb_type.values[0]                
-            if not check_residue_name.empty and pmb_type == pmb_type :              
+            mol_pmb_type = self.df.loc[self.df['name']==name].pmb_type.values[0]
+            if not check_residue_name.empty and mol_pmb_type == pmb_type:
```
jngrad, line 190: This is the bug revealed by Pylint.

Makefile
```diff
@@ -23,12 +23,16 @@ tests:
 	python3 testsuite/gcmc_tests.py
 	python3 testsuite/weak_polyelectrolyte_dialysis_test.py
 	python3 testsuite/globular_protein_tests.py
+
 sample:
-	python3 sample_scripts/peptide_simulation_example.py
+	python3 samples/peptide.py
 
 visual:
-	python3 handy_scripts/vmd-traj.py
```
jngrad, line 30: `make visual`  and `make sample` don't work.

jngrad (commented):
You can find a list of Pylint's rules [here](https://pylint.pycqa.org/en/latest/user_guide/checkers/features.html).

Due to the large number of changes, and the fact that some changes affect scripts that are not covered by the testsuite, I would recommend allocating more than 1 code reviewer to this PR.

pyMBE.py
```diff
@@ -60,7 +60,7 @@ def __init__(self, SEED, temperature=None, unit_length=None, unit_charge=None, K
         self.Kw=Kw*self.units.mol**2 / (self.units.l**2)
         self.units.define(f'reduced_energy = {self.kT}')
         self.units.define(f'reduced_length = {unit_length}')
-        self.units.define(f'reduced_charge = 1*e')
+        self.units.define('reduced_charge = 1*e')
         self.setup_df()
         return
```
jngrad, line 65: Regarding the usage of `eval()` later in the code, e.g. there and later in the code:
https://github.com/pyMBE-dev/pyMBE/blob/d10d17236217f71c2c13d8c7816e065c7d9e21e2/pyMBE.py#L1314

`eval()` is really dangerous. Flagged by pylint rule W0123. This is basically a backdoor (see Snyk's post about [code injection in Python](https://snyk.io/blog/code-injection-python-prevention-examples/)). Malicious actors can write a string in the pyMBE data.frame to execute arbitrary code with ease. This code prevents universities from creating a website to run pyMBE and let users interact with it (we did that for [ZnDraw](https://zndraw.icp.uni-stuttgart.de/)). This code might land pyMBE into a security vulnerability database (that happened to tqdm last week because of `eval()`: [CVE-2024-34062](https://www.cve.org/CVERecord?id=CVE-2024-34062) and [snyk-python-tqdm-6807582](https://security.snyk.io/vuln/SNYK-PYTHON-TQDM-6807582)).

Could this code be rewritten by serializing the Python `dict` returned by ESPResSo into a JSON string, and then deserializing the JSON string back to a Python `dict`? Caveat: some potentials might be a bit tricky, for example tabulated potentials might actually return numpy arrays instead of Python lists.

jngrad, line 65: Regarding the usage of `exit()` later in the code, there:
https://github.com/pyMBE-dev/pyMBE/blob/d10d17236217f71c2c13d8c7816e065c7d9e21e2/pyMBE.py#L2392-L2394

This should probably call `sys.exit(1)`. Flagged by pylint rule R1722. Right now, when a "hard check" fails, it doesn't fail hard; instead the Python script stops with exit code 0, which is interpreted as a success by the shell script. The shell script then continues its execution.

I am not familiar with the original design decision regarding module imports, so I didn't modify that line and din't add that rule to the pylint configuration file. Currently we import modules inside the pyMBE class, and we have to refer to them via `self`, e.g. `self.sys.exit(1)` instead of `sys.exit()`. I'm not really sure if there is a benefit to importing modules this way. If the reason was to defer module import until the pyMBE object is created, then this approach doesn't actually work, because import statements inside a class are not deferred (add a print statement inside the pyMBE class body to convince yourself); imports can only be deferred inside a class method or free function. If the reason was to make these modules available via the `pmb.pd` syntax, then I would recommend importing pandas in the script instead. Then one can put these import statements at the top of the pyMBE.py file.

testsuite/bond_tests.py
jngrad: Also, while changing line endings, one could also apply automatic formatting with `autopep8` to remove trailing whitespace characters. One would have to override `autopep8`'s max line length to avoid wrapping the code to 80 characters per line.

Makefile
pm-blanco: thank you for solving this bug!

pyMBE.py
pm-blanco: @jngrad I am pretty sure that we can rewrite the lines of code using `eval()` to solve this issue. We currently only support harmonic and FENE bonds anyway so it should not be that complicated. I was not convinced by the current implementation anyway. Should we address this issue in this PR or should we migrate it to a dedicated PR?

pm-blanco: @jngrad I think that we can move the module imports to the top of the `pyMBE.py` file, there was no good reason of why I decided to import them inside the pyMBE class other than it was the decision I took when I started developing the library and nobody actually pointed this out before :smile:

jngrad: This should probably be done in this PR. Then we can add `eval-used, #W0123` to the new file `.pylintrc` to prevent `eval()` from ever appearing in the codebase again.

I'm not super familiar with how the pyMBE data.frame is managed internally, though. This part of the code adds a new row using a Python dict serialized as a JSON string, but in the eval statement we convert the new value from a string to a Python object. If I use JSON to deserialize it back to a dict, the next time I call this function, the original value will be deserialized again even though it's already a Python object. I don't totally understand what this code does. Maybe the original author relied on the fact that `str` applied on a string is a no-op. Do we want this cell to always contain a dict? If so, why do we store it a string when adding it to the data.frame?

pm-blanco: @jngrad this is precisely why I was not convinced of the current implementation, I also do not understand why it is needed to do this `dict` to `str` back and forth transformation. We were discussing about it with @paobtorres when she implemented it and I know that it was due to some issue when parsing back the dataframe, but I would prefer to simply store it as a string that could be parsed by JSON for example.

Makefile
jngrad: @davidbbeyer can you please check these two Makefile targets? They still break on my workstation after applying the bug fixes.

pyMBE.py
jngrad: Alright! I moved them to the top of the file in f116cdc46b4b.

Makefile
pm-blanco: @jngrad @davidbbeyer  I solved the issue in `samples\peptide.py`, it was actually broken since #24. `make visual` works well in my work station...

jngrad: Here is what I get:
```
$ make visual
python3 visualization/vmd-traj.py
Loading frame ./frames/trajectory0.vtf
...
Loading frame ./frames/trajectory105.vtf
  File "/work/jgrad/pyMBE/visualization/vmd-traj.py", line 189, in <module>
    at_type=int(at_list[-1])
ValueError: invalid literal for int() with base 10: 'nH'
```

pm-blanco: Looks to me that is trying to load some broken trajectory file, how did you produce those frames? Did the simulation finish or did you stop it before finishing it?

jngrad: Ouch, I didn't realize it was reading frames created by `samples/peptide.py` :smiley: I erased those files and regenerated them by running the fixed `samples/peptide.py` script, and now it works.

lib/create_cg_from_pdb.py
```diff
@@ -1,9 +1,10 @@
-import numpy as np
-import pandas as pd 
+import sys
```
pm-blanco, line 1: @jngrad @paobtorres  I am not completely sure where is the issue but this script has changed its output. The following line
```bash
python3 lib/create_cg_from_pdb.py --download_pdb 1F6S --model 2beadAA --chain A
```
should create the same coarse-grained structure than the one we have stored in `parameters/globular_proteins/1f6s.vtf`

However, now it produces something different.

Original:
```
atom 1 name n resname n resid 1 chain A radius 0.7
atom 2 name E resname GLU resid 1 chain A radius 1.3735
atom 3 name CA resname GLU resid 1 chain A radius 0.7
atom 4 name CA resname GLN resid 2 chain A radius 0.7
atom 5 name Q resname GLN resid 2 chain A radius 1.4504
```
Current:
```
atom 1 name n resname n resid 1 chain A radius 0.7
atom 2 name E resname GLU resid 1 chain A radius 1.3735
atom 3 name CA resname GLU resid 1 chain A radius 0.7
atom 4 name Q resname GLN resid 2 chain A radius 1.4504
atom 5 name CA resname GLN resid 2 chain A radius 0.7
```
Note that the order of atom 5 and 4 have been reversed. The resulting structure looks the same as before by visual inspection in VMD but when I ran  the script in  `samples/Beyer2024/globular_protein.py` with the new structure, it breaks:
```bash
python3 samples/Beyer2024/globular_protein.py --pdb 1f6s --pH 2 --path_to_cg parameters/globular_proteins/1f6s.vtf --mode test
  File "/home/pablo/Desktop/pyMBE/samples/Beyer2024/globular_protein.py", line 164, in <module>
    pmb.create_protein(name=protein_name,
  File "/home/pablo/Desktop/pyMBE/pyMBE.py", line 1053, in create_protein
    particle_id = self.create_particle(name=residue_name,
  File "/home/pablo/Desktop/pyMBE/pyMBE.py", line 955, in create_particle
    raise ValueError(f"{name} must correspond to a label of a pmb_type='particle' defined on df")
ValueError: Ca must correspond to a label of a pmb_type='particle' defined on df
```
`create_cg_from_pdb.py`  also throws the following error, which I am not sure if it was before:

```bash
home/pablo/Desktop/pyMBE/lib/create_cg_from_pdb.py:282: FutureWarning: 
The provided callable <built-in function sum> is currently using SeriesGroupBy.sum. 
In a future version of pandas, the provided callable will be used directly.
 To keep current behavior pass the string "sum" instead.
```

pm-blanco (changes requested):

lib/create_cg_from_pdb.py
jngrad: I get this script to fail on the main branch too:
```
$ python3 samples/Beyer2024/globular_protein.py --pdb 1f6s --pH 2 --path_to_cg parameters/globular_proteins/1f6s.vtf --mode test
  File "/work/jgrad/pyMBE/samples/Beyer2024/globular_protein.py", line 163, in <module>
    pmb.create_protein(name=protein_name,
  File "/work/jgrad/pyMBE/pyMBE.py", line 1028, in create_protein
    particle_id = self.create_particle(name=residue_name,
  File "/work/jgrad/pyMBE/pyMBE.py", line 930, in create_particle
    raise ValueError(f"{name} must correspond to a label of a pmb_type='particle' defined on df")
ValueError: Ca must correspond to a label of a pmb_type='particle' defined on df
```
Printing the values that already exist in the data frame column reveals the following:
```
['CA' 'n' 'E' 'Q' 'L' 'T' 'K' 'C' 'V' 'F' 'R' 'D' 'Y' 'S' 'P' 'W' 'H' 'A'
 'I' 'N' 'M' 'c' 'AA-n' 'AA-E' 'AA-Q' 'AA-L' 'AA-T' 'AA-K' 'AA-C' 'AA-V'
 'AA-F' 'AA-R' 'AA-D' 'AA-Y' 'AA-S' 'AA-P' 'AA-W' 'AA-H' 'AA-A' 'AA-I'
 'AA-N' 'AA-M' 'AA-c' '1f6s' 'Na' 'Cl']
```
Could this be an issue with case-sensitive names? "CA" may refer to either Ca<sup>2+</sup> or C-alpha carbons, both of which are present in PDB id 1f6s.

Regarding the `FutureWarning`, I have no opinion, since I'm not familiar with this specific part of the pyMBE code. I would be surprised if this was introduced by my changes, since in that function I only adapted the indentation level to 4-characters instead of 8-characters.

tutorials/pyMBE_tutorial.ipynb
```diff
@@ -1625,14 +2374,122 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 36,
```
jngrad, line 2377: Please clear output cells before committing notebook changes :)

pm-blanco: ups! I thought I did clean it up, but probably I forgot to save it afterwards...

lib/create_cg_from_pdb.py
pm-blanco: You are absolutely right @jngrad that was the issue. I checked and for that structure one needs to run the script providing the name of the metal ion 
```bash
python3 samples/Beyer2024/globular_protein.py --pdb 1f6s --pH 2 --path_to_cg parameters/globular_proteins/1f6s.vtf --mode test  --metal_ion_name Ca --metal_ion_charge 2
```
with this line the script runs both in the main branch and this PR.

I am still a bit confused with the change in output of  `lib/create_cg_from_pdb.py`, but it seems harmless as the CI tests passes with the new structure. In the future, when we continue developing the features for globular proteins we need a more robust way of dealing with metal ions and other heterogenous elements in PDB.

paobtorres: @pm-blanco @jngrad  I tested both `create_cg_from_pdb.py` and `globular_protein.py` (after creating a new cg model of 1f6s) and both worked okay. 
Regarding the `FutureWarning` I believe its something that started to show up since the later version of pandas.

pm-blanco (dismissed):

pm-blanco:
I have been checking around, testing the parts that are not checked by the CI and I did not find any further issue with the PR. Still, it would be good if at least @paobtorres or @davidbbeyer  checks as well before I merge this PR

lib/create_cg_from_pdb.py
jngrad: > we need a more robust way of dealing with metal ions and other heterogenous elements in PDB

Looking at [Table 1](https://pubs.acs.org/doi/10.1021/acs.jproteome.0c00906#tbl1) in [10.1021/acs.jproteome.0c00906](https://doi.org/10.1021/acs.jproteome.0c00906), we might as well add the complete periodic table in pyMBE :smile: 

One could also parse the PDBx metadata in CIF files to list metallic ions, and when relevant (e.g. iron) their oxidation state:

* [1F6S.cif](https://files.rcsb.org/header/1F6S.cif)
   ```
   1 polymer     nat ALPHA-LACTALBUMIN 14202.048 6  ? ? ? ? 
   2 non-polymer syn 'CALCIUM ION'     40.078    6  ? ? ? ? 
   3 water       nat water             18.015    89 ? ? ? ? 
   ...
   CA  non-polymer         . 'CALCIUM ION'   ? 'Ca 2'           40.078  
   ```
* [5ZTA.cif](https://files.rcsb.org/header/5ZTA.cif)
   ```
   1 polymer     man 'Sirohydrochlorin ferrochelatase' 30237.533 2  4.99.1.4 ? ? ? 
   2 non-polymer syn 'FE (III) ION'                    55.845    2  ?        ? ? ? 
   3 water       nat water                             18.015    13 ?        ? ? ? 
   ...
   FE  non-polymer         . 'FE (III) ION'  ? 'Fe 3'           55.845  
   ```

samples/Beyer2024/create_paper_data.py
```diff
@@ -51,7 +52,7 @@
         raise RuntimeError()
     pH_range = np.linspace(2, 12, num=21)
     for pH in pH_range:
-        run_command=["python3", script_path, "--sequence", sequence, "--pH", str(pH), "--mode", mode]
+        run_command=[sys.executable, script_path, "--sequence", sequence, "--pH", str(pH), "--mode", mode]
```
jngrad, line 55: Just in case we run on a platform where `python` is used instead of `python3`. Quite rare, but can happen depending on how a virtual environment is configured, or how Python gets packaged on macOS.

Makefile
```diff
@@ -3,36 +3,38 @@
...
+	${PYTHON} testsuite/cph_ideal_tests.py
+	${PYTHON} testsuite/grxmc_ideal_tests.py
+	${PYTHON} testsuite/peptide_tests.py
+	${PYTHON} testsuite/gcmc_tests.py
+	${PYTHON} testsuite/weak_polyelectrolyte_dialysis_test.py
+	${PYTHON} testsuite/globular_protein_tests.py
 
 sample:
-	python3 samples/peptide.py
+	${PYTHON} samples/peptide.py
```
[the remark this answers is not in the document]
jngrad: For more context: Makefiles variables can be overridden by user-defined variables with the same name passed after the target name. You are probably already familiar with this technique, as it is often used when tinkering with software that refuses to build on a cluster. For example with `make VERBOSE=1` in the ESPResSo build folder, you will see which C++ compiler commands are being executed (add an empty line in any C++ file if needed to make the build out-of-date).

lib/create_cg_from_pdb.py
paobtorres: @jngrad @pm-blanco  I think this is a great idea! I can do it, because at least for proteins we should expand the metal ions supported. We can later discuss more about adding this.

Makefile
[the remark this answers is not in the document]
pm-blanco: having code coverage would be of great help to continue developing the CI tests!!

pm-blanco (dismissed):

samples/Beyer2024/create_paper_data.py
```diff
@@ -40,7 +41,7 @@
 labels_fig7=["7a", "7b", "7c"]
 
 if fig_label in labels_fig7 and not plot:
-    script_path=pmb.get_resource(f"samples/Beyer2024/peptide.py")
+    script_path=pmb.get_resource("samples/Beyer2024/peptide.py")
```
paobtorres, line 44: @jngrad @pm-blanco  It's not related to the changes on the PR but we need to fix this. In line 43, Is there a reason for `and not plot`? Because if we run the script with `--plot` the `sequence` of the peptides is not defined on that section (line 183) and the script crashes.

jngrad: I am not too familiar with this script, so I would trust @pm-blanco's opinion here.

Looking at the PR diff again, I also just stumbled upon the following oddity:
https://github.com/pyMBE-dev/pyMBE/blob/d10d17236217f71c2c13d8c7816e065c7d9e21e2/pyMBE.py#L1859-L1860
I'm pretty confident `sys.exit(1)` would be more appropriate than raising an exception. This would also be more consistent with other instances of `if hard_check:` in the code. Right now, it's unclear to me how a pyMBE simulation is supposed to react when that exception is raised, because it won't be caught by `except Exception` (since it derives from `BaseException`), and it sends a signal as a side-effect, which propagates through subprocess calls in non-trivial ways. For example, subprocess can convert it from a signal to an exception! Consider the following example:
```python

>>> import subprocess
>>> subprocess.check_output(["python3", "-c", "raise ValueError()"])
subprocess.CalledProcessError: Command '['python3', '-c', 'raise ValueError()']' returned non-zero exit status 1.
>>> subprocess.check_output(["python3", "-c", "raise KeyboardInterrupt()"]) # signal becomes and exception!
subprocess.CalledProcessError: Command '['python3', '-c', 'raise KeyboardInterrupt()']' died with <Signals.SIGINT: 2>.
```

The [`KeyboardInterrupt`](https://docs.python.org/3/library/exceptions.html#KeyboardInterrupt) documentation has a disclaimer to deter people from using it in the way we do.

pyMBE.py
```diff
@@ -60,7 +80,7 @@ def __init__(self, SEED, temperature=None, unit_length=None, unit_charge=None, K
         self.Kw=Kw*self.units.mol**2 / (self.units.l**2)
         self.units.define(f'reduced_energy = {self.kT}')
         self.units.define(f'reduced_length = {unit_length}')
-        self.units.define(f'reduced_charge = 1*e')
+        self.units.define('reduced_charge = 1*e')
```
jngrad: Hey! The reduced charge should be `f'reduced_charge = {unit_charge}'`, not a hardcoded unit of elementary charge. Right now, the `unit_charge` provided by the user is silently ignored. Also, this block of code that defines defaults is duplicated from `set_reduced_units()`, which we could call instead. Please check c7fff1658f2.

davidbbeyer (commented):
Apart from the small points raised above, it looks good to me.

samples/peptide.py
```diff
@@ -39,7 +34,7 @@
 
 # Peptide parameters
 
-sequence = 'nEEEEEc'
+sequence = 'EEEEDDDD'
```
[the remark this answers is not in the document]
pm-blanco: @davidbbeyer the reason is that the termini were actually not parametrized in `parameters/peptides/Lunkad2021.json`. In earlier versions of this script, this was bypassed by adding a generic parametrization for the non-parametrized aminoacid particles. This was a convenient solution to allow to run any peptide sequence, but not really a good practice (since we were basically teaching people to mix parameters instead of thinking about what should be a good parametrization). With @mariusaarsten we have done a more complete parametrization of the amino acids which he will add in his master thesis and we will also provide in pyMBE so we can have a first complete set of parameters for all aminoacids.

maintainer/standarize_data.py
```diff
@@ -59,7 +58,6 @@
     data=np.loadtxt(ref_path, delimiter=",")
     Z_ref=data[:,1]         
     Z_ref_err=data[:,2]
-
     pH_range = np.linspace(2, 12, num=21)
```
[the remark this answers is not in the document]
pm-blanco: solved!

pyMBE.py
pm-blanco: @jngrad good catch!! thank you for fixing this

pm-blanco (approved):
@davidbbeyer @paobtorres I think we are ready now for merging this PR. If you agree, please leave a review approving these changes (go to the files changes tab, review changes botton, and press approve). @paobtorres please open a dedicated PR to fix the issues that you spotted in the `create_paper_data.py` script.

PR #62, opened 2024-05-17
analysis tests

kosovan:
Provides unit tests for:

* [x] `lib.analysis.get_params_from_file_name()`
* [x] `lib.analysis.block_analyze()`
* [x] `lib.analysis.get_dt()`
* [x] `lib.analysis.split_dataframe_in_equal_blocks()`
* [x] `lib.analysis.split_dataframe()`
* [x] `lib.analysis.add_data_to_df()`
* [x] `lib.analysis.analyze_time_series()`

Deprecates:
* [x] `lib.analysis.do_binning_analysis()`
* [x] `lib.analysis.merge_time_series_dfs()`
* [x] `lib.analysis.get_time_series_from_average_df()`
* [x] `lib.analysis.read_csv_file()`
* [x] `lib.analysis.get_distribution_from_df()`
* [x] `lib.analysis.create_histogram_df_from_distribution_list()`
* [x] `lib.analysis.find_index_with_value_in_df()`

Coverage in `analysis.py`: 100%; Increases the total coverage of the module to 83%


kosovan:
@pm-blanco Are these analysis functions actually used by pyMBE? Maybe we should consider removing or at least updating some of them?

pm-blanco:
@kosovan `block_analyze` and functions called therein are currently called by the sample scripts `samples/branched_polyampholyte.py`, `samples/peptide_mixture_grxmc_ideal.py`, `samples/peptide.py`  and also by the script that we use to reproduce the data of our manuscript , `samples/Beyer2024/create_paper_data.py` , via `analysis.analyze_time_series()`. I am not sure if all functions in `lib.analysis` are actually called within pyMBE, it would definitively by worth checking if we can clean-up some. Updating them to their current counterpart in our private repository is a good idea, would you take care of that?

pm-blanco:
@kosovan I took the liberty to finish this PR. I cleaned up `lib/analysis.py` of all the functions that are not used within pyMBE and I updated the functions to work following our current standards. I have also provided unit tests for all the functions in the library, which cover the 100% of the code.

pm-blanco:
@paobtorres this PR is now ready for review

lib/analysis.py
paobtorres, line 1: There are a few inconsistencies in the docstring. For example, in Args from `analyze_time_series()` has no items (-) and the Returns provides only the format output. Then, in `block_analyze()` each Args has an item symbol (-) the the Returns contains the variable name of the output.

paobtorres (changes requested):
I think the docstring in `analysis.py` should be changed to keep a consisting format. Other than that, I did some testing and everything ran successfully.

lib/analysis.py
pm-blanco: @paobtorres thank you for your revision and for bringing up this point. I have reviewed the docstrings of `analysis.py` and `pyMBE.py` to try to standardize the format and, in the process, I solved numerous inaccuracies and ambiguities on variable types. Regarding the format of the Return arguments, I think that we should actually not enforce that there should be a variable name in the docstrings because there are some functions like `pmb.check_if_df_cell_has_a_value` or `check_if_name_is_defined_in_df` that do not really store their output into a variable but simply return it and then the user can put any arbitrary variable name to it. I would suggest to either drop all variable names for return arguments (because they are actually not important for users anyway) or to only add the variable name in return arguments when it is actually defined in the function.

paobtorres: @pm-blanco I think that depends on which Python code docstring guideline we follow. I believe that when there is a variable name we should put it because it helps the user to better understand how each function works by reading the docstring.

pm-blanco: @paobtorres we follow the [Google Style Python Docstrings](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html). In that standard, the variable names is not provided for return arguments. I do not think that it is enforced that we should not provide variable names, so we can agree that we provide variable names for return variables to help users reading the code but we do not enforce that all functions should have a return variable name.

paobtorres (approved):
I reviewed the changes and quickly ran a test everything is ok.

PR #65, opened 2024-05-30
add admin team to authors list

pm-blanco:
Adds the current admin team to `AUTHORS.md` to clarify the roles in our software.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -1,12 +1,14 @@
 pyMBE is maintained and developed by several authors from different institutions in collaboration.
 The following people have contributed to pyMBE:
 
-## Core developers
-- David Beyer (University of Stuttgart, Germany)
+## Admins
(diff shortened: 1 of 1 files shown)


pm-blanco:
@jngrad should we use ESPResSo/4.2.2 for the first release of pyMBE? The CI is not passing anymore due to the warning of the new available version of espresso

jngrad:
> @jngrad should we use ESPResSo/4.2.2 for the first release of pyMBE? The CI is not passing anymore due to the warning of the new available version of espresso

I would recommend updating to 4.2.2 in CI as part of our first release, since it contains relevant bugfixes for Monte Carlo. Although, our CI pipelines should not fail when using older ESPResSo releases, because that would be quite disruptive to open PRs. I have relayed the issue to the EESSI team and they are working out a fix in https://github.com/EESSI/software-layer/pull/589.

jngrad (approved):

PR #68, opened 2024-06-04
Copyright

### LINKED ISSUE #53: Add copyright notice to source files.
@pm-blanco Like it is done for instance also in ESPResSo, we should add a notice regarding the copyright and exclusion of warranty to _every_ source file. The following is suggested by the Free Software Foundation:

> \<one line to give the program's name and a brief idea of what it does.\>
> Copyright (C) \<year\>  \<name of author\>
> 
> This program is free software: you can redistribute it and/or modify
> it under the terms of the GNU General Public License as published by
> the Free Software Foundation, either version 3 of the License, or
> (at your option) any later version.
> 
> This program is distributed in the hope that it will be useful,
> but WITHOUT ANY WARRANTY; without even the implied warranty of
> MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
> GNU General Public License for more details.
> 
> You should have received a copy of the GNU General Public License
> along with this program.  If not, see <http://www.gnu.org/licenses/>.

A corresponding note should also be displayed on the CLI if pyMBE is used:

> \<program\>  Copyright (C) \<year\>  \<name of author\>
> This program comes with ABSOLUTELY NO WARRANTY; for details type \`show w'.
> This is free software, and you are welcome to redistribute it
> under certain conditions; type \`show c' for details.

pinedaps:
Fixes #53 

List of changes:

I added the license header to all Python/Cython/Bash/C++/CUDA/Doxygen files in the pyMBE repository.

The template I used can be found at '/maintainer/header_template_license.txt', so anyone can copy and paste it when adding new Python/Cython/Bash/C++/CUDA/Doxygen files. However, we need to remember to update the year in subsequent years.

### DIFF
--- a/maintainer/header_template_license.txt
+++ b/maintainer/header_template_license.txt
@@ -1,17 +1,17 @@
-#pyMBE : The python-based Molecule Builder for ESPResSo.
-#Copyright (C) 2024 pyMBE-dev team  
 #
-#This file is part of pyMBE.
+# Copyright (C) 2024 pyMBE-dev team  
 #
-#pyMBE is free software: you can redistribute it and/or modify
-#it under the terms of the GNU General Public License as published by
-#the Free Software Foundation, either version 3 of the License, or
-#(at your option) any later version.
+# This file is part of pyMBE.
 #
-#pyMBE is distributed in the hope that it will be useful,
-#but WITHOUT ANY WARRANTY; without even the implied warranty of
-#MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#GNU General Public License for more details.
+# pyMBE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
 #
-#You should have received a copy of the GNU General Public License
-#along with this program.  If not, see <http://www.gnu.org/licenses/>.
+# pyMBE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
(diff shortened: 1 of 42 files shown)


maintainer/header_template_license.txt
```diff
@@ -0,0 +1,17 @@
+#pyMBE : The python-based Molecule Builder for ESPResSo.
+#Copyright (C) 2024 pyMBE-dev team
```
jngrad, line 2: I would recommend introducing a whitespace between the hashtag and text, if the line contains text. If we ever add a code formatter like autopep8 or Black, they will complain at the missing whitespace.

Regarding the first line, I don't remember ever seeing a license header file featuring a program description above the copyright year. I don't mind it, but I would introduce an extra line between `pyMBE` and `Copyright`, so that the copyright year range belongs to its own paragraph. This is actually the most important part of the license header, as you have to copy-paste it when re-using code, and thus needs to be immediately visible. See for example [espressomd/espresso@`abb087c`:maintainer/walberla_kernels/walberla_lbm_generation.py](https://github.com/espressomd/espresso/blob/abb087c7027b926687efb53a5ac9630d53488923/maintainer/walberla_kernels/walberla_lbm_generation.py#L2-L3)

For the notebook, I'm not sure if we have to include the license header in a Markdown cell. While Jupyter notebooks are text files, they are not really supposed to be opened in a text editor by users, and the official JSON schema doesn't officially support a license field (only basic metadata fields like title and author). As such, I would treat it as any binary file, which inherit the license of the entire project (GNU GPL v3).

Maybe @pm-blanco can tell us more after meeting with the NTNU archivist. Software projects with an end-user license agreement (EULA) sometimes write in their EULA something like "[name] is free software; you can redistribute it and/or modify it under the terms of [free software license]. This license applies to all files distributed in the project project, including all source code, binaries and documentation, unless otherwise stated in the file." But I'm not sure if we can write something to that effect in our readme file, because open source software licenses are not exactly an EULA: they regulate the *redistribution* of software by end users, not its *usage* by end users. This is a question we will have to answer in the pyMBE logo, whose licensing terms depend on the service that was used to help its generation (I would imagine it is in the public domain). The SVG file format has metadata fields to indicate the license.

pinedaps: I can add the white spaces and regarding the first line, that appears in the license template described in issue #53. 
Further changes in tutorials can be discussed tomorrow in the meeting.

pm-blanco: I would simply drop the first line. I have requested an appointment with the library publishing team at NTNU but they have not replied to me yet. For now, maybe I would not put the copyright in the notebooks.

jngrad (approved):

PR #76, opened 2024-07-15
Introduce author tier system

pm-blanco:
Introduces a tier system to reward contributors to pyMBE, establishes:
- What are considered valid contributions to the library.
- The criteria for classifying a contributor Tier-0 or Tier-1.
- The rewards for each tier. 

@kosovan @jngrad  please let me know if you agree with this framework


jngrad (commented):
Looks good to me. I would probably replace "carpentry" by "architecture", since the latter is more common.

kosovan (approved):
I think it is final.

PR #78, opened 2024-08-05
np.NaN deprecated for np.nan

Zitzeronion:
Running the tutorial with numpy v2.0 yields an error due to `np.NaN`. Both `np.NaN` and `np.nan` have been used, changed the remaining two NaN to nan.


jngrad (approved):
LGTM

PR #80, opened 2024-08-07
Code duplication

Zitzeronion:
In the second peptide code cell `pmb.load_interaction_parameters(filename = pmb.get_resource('parameters/peptides/Lunkad2021.json'))` appeared twice.


pm-blanco (approved):
Looks good to me, thank you.

PR #81, opened 2024-08-07
Folder name `sample_scripts` changed to `samples`

Zitzeronion:
There is no folder anymore which is called `sample_scripts` however there is a folder called `samples` which has a `peptide.py` file.


pm-blanco (approved):
Looks good to me. Thank you!

PR #84, opened 2024-08-15
Require Python dependency versions compatible with ESPResSo 4.2

### LINKED ISSUE #83: numpy v2.0 error with espresso v4.2.2
When running [peptide.py](samples/peptide.py) with numpy 2.0 and espresso v4.2.2 there seems to be an issue. 
```shell
python3 samples/peptide.py 
  File ".../samples/peptide.py", line 21, in <module>
    import espressomd
  File ".../espresso-v422/build/src/python/espressomd/__init__.py", line 21, in <module>
    from . import _init
  File "_init.pyx", line 1, in init espressomd._init
ValueError: numpy.dtype size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject
```

There is a fix which can be found on [stackoverflow](https://stackoverflow.com/questions/78650222/valueerror-numpy-dtype-size-changed-may-indicate-binary-incompatibility-expec) that basically suggests to downgrade the numpy version with 

`(pymbe) pip install "numpy<2"`

Did not run into this problem with espresso v4.3 and numpy v2.0

jngrad:
Fixes #83

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,8 +1,8 @@
-numpy>=1.23
+numpy>=1.23,<2.0
 pandas>=1.5.3
 pint>=0.20.01
 pint-pandas>=0.3
 biopandas==0.5.1.dev0
-scipy
-matplotlib
-tqdm
+scipy>=1.8.0
+matplotlib>=3.5.1
+tqdm>=4.57.0


pm-blanco (approved):
Looks good to me!

Issue #60, opened 2024-05-17
Remove global state


jngrad:
Global variables are notorious for introducing [race conditions](https://en.wikipedia.org/wiki/Race_condition), which can cause bugs that are hard to reproduce and investigate. They also have shared ownership with multiple objects, thus client code cannot make assumptions about their state.

To illustrate pitfalls from global variables, consider the following example:

```python
class pymbe_library():
    import pint
    units = pint.UnitRegistry()
    SEED = None
    def __init__(self, SEED):
        self.SEED = SEED
        unit_length = 0.355 * self.units.nm
        self.units.define(f"reduced_length = {unit_length}")
```

This is a simplified version of the pyMBE class that was available up to and including d10d17236217f71c2c13d8c7816e065c7d9e21e2. The class members `units` and `SEED` are bound to the class, not to class instances. You can access them from outside with `print(pymbe_library.SEED)`.

Here is a MWE:

```python
pmb1 = pymbe_library(SEED=42)
pmb2 = pymbe_library(SEED=43)
print(f"{pmb1.SEED=}")
print(f"{pmb2.SEED=}")
print(f"{pmb1.units=}")
print(f"{pmb2.units=}")
print(id(pmb1.units) == id(pmb2.units))
```
Output:
```
pmb1.SEED=42
pmb2.SEED=43
pmb1.units=<pint.registry.UnitRegistry object at 0x7235c663fa0>
pmb2.units=<pint.registry.UnitRegistry object at 0x7235c663fa0>
True
```

The `SEED` members are different, but the `unit` members are shared between the two objects (same memory address). These members differ in behavior because integers are immutable, while Pint objects are mutable ([quick refresher](https://realpython.com/python-mutable-vs-immutable-types/)). The rules for an object's mutability and scope can be a bit difficult to grasp, and while Python developers should know about them, we cannot expect them to know how the pyMBE class constructor uses these rules internally to manage the lifetime and ownership of its class members. I would advocate for the [principle of least surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) by making these class members fully owned by the class instance, i.e. the constructed object, rather than by the class itself.

This can be achieved without any API change with the following rewrite:

```python
import pint
class pymbe_library():
    def __init__(self, SEED):
        self.SEED = SEED
        self.units = pint.UnitRegistry()
        unit_length = 0.355 * self.units.nm
        self.units.define(f"reduced_length = {unit_length}")
```
[...]

PR #86, opened 2024-08-26
Use 2019 SI base units

jngrad:
The SI base units have been redefined as natural constants with exact precision in 2019.

### DIFF
--- a/lib/handy_functions.py
+++ b/lib/handy_functions.py
@@ -19,32 +19,31 @@
 def setup_electrostatic_interactions (units, espresso_system, kT, c_salt=None, solvent_permittivity=78.5, method='p3m', tune_p3m=True, accuracy=1e-3,verbose=True):
     """
     Sets up electrostatic interactions in espressomd. 
-    Inputs:
-    system: instance of espressmd system class
-    c_salt: Added salt concentration. If provided, the program outputs the debye screening length. It is a mandatory parameter for the Debye-Huckel method. 
-    solvent_permittivity: Solvent relative permitivity, by default chosen per water at 298.15 K
-    method: method prefered for computing the electrostatic interactions. Currently only P3M (label = p3m) and Debye-Huckel (label = DH) are implemented
-    tune_p3m: If true (default), tunes the p3m parameters to improve efficiency
-    accuracy: desired accuracy for electrostatics, by default 1e-3
-    verbose (`bool`): switch to activate/deactivate verbose. Defaults to True.
+
+    Args:
+        units(`pint.UnitRegistry`): Unit registry.
+        espresso_system: instance of espressmd system class
+        kT(`float`): Thermal energy.
+        c_salt: Added salt concentration. If provided, the program outputs the debye screening length. It is a mandatory parameter for the Debye-Huckel method.
+        solvent_permittivity: Solvent relative permitivity, by default chosen per water at 298.15 K
+        method: method prefered for computing the electrostatic interactions. Currently only P3M (label = p3m) and Debye-Huckel (label = DH) are implemented
+        tune_p3m: If true (default), tunes the p3m parameters to improve efficiency
+        accuracy: desired accuracy for electrostatics, by default 1e-3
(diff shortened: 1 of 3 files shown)


jngrad:
The old units were experimentally determined and had fluctuating precision. Each textbook had a slightly different value.

We are currently using the old values, and they have a tiny deviation from the new values. For example, the ratio of pyMBE Boltzmann constant and the corresponding 2019 constant deviates from unity by three times the machine-precision of single-precision floating point values:

```python
import pyMBE, numpy as np, scipy.constants
eps = scipy.constants.k / pyMBE.pymbe_library.Kb.magnitude - 1.
print(eps / np.finfo(np.float32).eps)
```

For the electric charge, it's one order of magnitude smaller than machine precision.

I don't think it makes much of a difference in long simulations due to thermal fluctuations, but it may cause test failures in unit tests if the `atol` parameter of  `np.testing.assert_*()` is larger than epsilon, and the test uses the new value of kB. The 2019 value for kB is exact and identical on NIST ([link](https://physics.nist.gov/cgi-bin/cuu/Value?k)), IUPAC ([link](https://goldbook.iupac.org/terms/view/B00695)), BIPM ([link](https://www.bipm.org/documents/20126/41483022/SI-Brochure-9-EN.pdf)), and scipy.

pm-blanco:
@jngrad it looks to good to me. I ran the functional tests and everything seems to work fine. Should we add a small unit tests that checks that pyMBE uses the same fundamental constants as the ones in scipy, just to ensure that this does not break again in the future?

jngrad:
> Should we add a small unit tests that checks that pyMBE uses the same fundamental constants as the ones in scipy, just to ensure that this does not break again in the future?

Good point. I didn't realize the values of the constants were overridden in `set_reduced_units()`, yikes!
We *really* need to remove these global variables :sweat_smile: (#60)

Side-note: I personally find using `Kb` instead of `kB` for the Boltzmann constant $k_B$ to be a bit unfortunate considering how similar it is to the equilibrium constant $K_b$, and that we define the self-ionization constant $K_w$ in member `Kw`.

pm-blanco:
@jngrad 

> Good point. I didn't realize the values of the constants were overridden in set_reduced_units(), yikes!

Yes, the problem is that in `set_reduced_units()` one needs to create a new `pint.UnitRegistry()` object to redefine the reduced unit system. Then, one needs to basically overwrite all previously existing pint variables because it is not possible to do operation between pint objects defined with different unit registries. That is why one needs to overwrite the constants so one can still operate with them with the new unit system.

The Pint documentation is not very extensive regarding this definitions of new units [docs](https://pint.readthedocs.io/en/stable/advanced/defining.html). The solution that we have right now is not perfect because a user could still define particles with an old unit registry then set a different set of reduced units and then proceed to define new pyMBE objects with the new unit registry. If could simply redifine the reduced units without creating a new `pint.UnitRegistry()` object  that would be a much more robust solution.

> We really need to remove these global variables 😅 (https://github.com/pyMBE-dev/pyMBE/issues/60)
I agree, we should deal with this issue before the pyMBE 1.0.0 release. There is no major reason for having the global and we only need to have them defined by the class constructor instead.

> I personally find using Kb instead of kB for the Boltzmann constant k B to be a bit unfortunate considering how similar it is to the equilibrium constant K b , and that we define the self-ionization constant K w in member Kw.

I have not realized of this issue. You are right that it might be a bit missleading. Feel free to change the notation in this PR if you like.

jngrad:
> The solution that we have right now is not perfect because a user could still define particles with an old unit registry then set a different set of reduced units and then proceed to define new pyMBE objects with the new unit registry.

We could let the user provide the unit registry as an argument to the constructor. That would require users to write two extra lines to import pint and create their unit registry, but only if they don't use the default reduced units system. That would also make it clear that the unit registry belongs to the user, and pyMBE keeps a reference to it.

This is probably out of scope for this PR, though.

> Feel free to change the notation in this PR if you like.

Done.

PR #87, opened 2024-08-28
Switch to CTest

jngrad:
Replace the Makefile-based test driver by the CTest driver. Tests can now run in parallel with `make unit_tests -j8`.


jngrad:
The CTest tool is described in more detail in the the "Mastering CMake" book, chapter [*Testing With CMake and CTest*](https://cmake.org/cmake/help/book/mastering-cmake/chapter/Testing%20With%20CMake%20and%20CTest.html). The CMake community provides detailed instructions on [*Using CTest and CDash without CMake*](https://gitlab.kitware.com/cmake/community/-/wikis/doc/ctest/Using-CTEST-and-CDASH-without-CMAKE). I skipped the CDash part, since I don't have much experience with the dashboard and wasn't sure how it would integrate in GitHub Actions. Most pyMBE and ESPResSo users are already familiar with CTest, and its ASCII-based reporting should be sufficient for our needs.

Long tests are sorted by decreasing runtime. This way, the first time `make functional_tests` is called, the CTest scheduler runs the slowest tests first. This cuts the `samples` job runtime from 50 min down to 20 min (logfile [29390808749](https://github.com/jngrad/pyMBE/actions/runs/10604352476/job/29390808749)). This only works the first time, though! For end users who run `make functional_tests` or `make tests` multiple times, the test order may not match with the order specified in `testsuite/CTestTestfile.cmake` after the first run. That's because the CTest scheduler keeps track of test statistics in `testsuite/Testing/Temporary`, such as runtimes and test failures, and uses this information to improve future runs. For example, failed tests in the last run will be scheduled to start first the next run, so as to give users rapid feedback on whether their latest code changes have fixed the failing tests.

While this PR adds CMake as a dependency, I have no intention of making pyMBE a CMake project. CMake is required by CTest, but anyone can edit the `testsuite/CTestTestfile.cmake` without the need to be familiar with the CMake syntax. In fact, variables like [`CMAKE_CURRENT_SOURCE_DIR`](https://cmake.org/cmake/help/latest/variable/CMAKE_CURRENT_SOURCE_DIR.html) are not even available, since we don't have a CMake configuration stage. This can come as a surprise to experienced CMake users.

testsuite/henderson_hasselbalch_tests.py
```diff
@@ -16,87 +16,97 @@
...
+                                pH_list = pH_range)
+        Z_HH_2 = pmb.calculate_HH(molecule_name = "peptide_2",
+                                pH_list = pH_range)
 
-print("*** Test passed ***\n")
+        """
+        with open(self.data_root / "HH.csv", "wb") as f:
+            np.savetxt(f, np.asarray(Z_HH_1).reshape(1,-1), delimiter=",")
+            np.savetxt(f, np.asarray(Z_HH_2).reshape(1,-1), delimiter=",")
+        """
```
pm-blanco, line 67: Could you please remove this comment?

testsuite/henderson_hasselbalch_tests.py
```diff
@@ -16,87 +16,97 @@
...
+                           "peptide_2": pep2_concentration},
+                c_salt = c_salt,
+                pH_list = pH_range)
 
-print("*** Test passed ***\n")
+        """
+        with open(self.data_root / "HH_Donnan.csv", "wb") as f:
+            np.savetxt(f, np.asarray(HH_Donnan_dict["charges_dict"]["peptide_1"]).reshape(1,-1), delimiter=",")
+            np.savetxt(f, np.asarray(HH_Donnan_dict["charges_dict"]["peptide_2"]).reshape(1,-1), delimiter=",")
+        """
```
pm-blanco, line 89: Could you please remove this comment?

requirements.txt
```diff
@@ -6,3 +6,4 @@ biopandas==0.5.1.dev0
 scipy>=1.8.0
 matplotlib>=3.5.1
 tqdm>=4.57.0
+cmake>=3.22.1
```
pm-blanco, line 9: I think we should add a comment here to note that this dependency is only required for testing, in case it might be an issue for some specific systems of our users

requirements.txt
```diff
@@ -6,3 +6,4 @@ biopandas==0.5.1.dev0
 scipy>=1.8.0
 matplotlib>=3.5.1
 tqdm>=4.57.0
```
pm-blanco, line 8: Maybe we should add a note on `tqdm`, it is only a dependency because we use it in some of our sample scripts but it is not really a dependency of pyMBE as a module

pm-blanco (changes requested):
Thank you very much for this PR and the useful links @jngrad! I did some testing in my office computer and the new framework for tests seems to work very well and is way faster than before, which will save both on our Github quota and on the time spending debugging code. I also think that it should be clear from your notes how to add new tests with the new framework, but @paobtorres  should take a look into this PR since she is the one that is going to develop tests in the near future.

I think that we should note in the requierements that `cmake` is a dependency only for testing. Since we are working in a virtual environment it should in principle not be a problem to have it installed by default, but just in case it might be a potential issue for some of our users. I added a similar note on `tqdm` which is actually only needed because we use it in some of our sample scripts

requirements.txt
jngrad: Can do.

As a general comment, `requirements.txt` isn't the right tool to document the dependencies of two different audiences, in this case end users vs. developers. I wasn't even sure if we should even mention CMake at all here, since `requirements.txt` is used by PyPI to document user requirements, and they will never have to call CTest or Pylint when running pyMBE simulations. But we still need to document CMake and Pylint version requirements somewhere for our contributors. We could do that in the readme file, but this can quickly become outdated.

This is a common issue for Python package maintainers, and I don't have a satisfying solution at hand. Maintaining two `requirements.txt` files is discouraged due to the cybersecurity threat it poses (someone once registered a faulty Python package [`requirements-dev.txt`](https://pypi.org/project/requirements-dev/) as a joke to raise awareness, although now this package raises a meaningful error message). In the ESPResSo user guide we use the requirement file as a constraint file with syntax `pip3 install -c requirements.txt numpy scipy jupyterlab` instead of `-r`, but then you need to specify the subset of Python packages you want to install.

I also quickly went through the [`poetry`](https://python-poetry.org/) tutorial, since it was recommended to me as a potential solution, and while it solves the issue of sorting packages into hard and soft dependencies using a tree structure with optional leaves, under the hood every change to the text file triggers and installation of the corresponding package in a `virtualenv` environment. That's because poetry is primarily designed for packaging and publishing projects, which is a lot more than what we are asking for, and can be a bit confusing to contributors who are only interested in quickly setting up a `venv` or `conda` environment. I see a lot of value in poetry, however I am not fully convinced this is the right tool for us, at this specific point in time.

Here is a minimal working example I ran in a Docker container; the `@^1.23` and `@~7.4.4` labels generate `>=1.23,<2.0` resp. `>=7.4.4,<7.5` version ranges, but these are processed internally to yield a pip input file with frozen version numbers that satisfy these ranges:

```sh
pip3 install --user poetry poetry-plugin-export
poetry new pyMBE
cd pyMBE
poetry add --group dev 'pylint@3.0.3' 'coverage@~7.4.4' 'cmake@^3.22.1'
poetry add --group docs 'pdoc@^14.3'
poetry add 'numpy@^1.23'
poetry lock
poetry export -f requirements.txt --without-hashes --only dev -o only-dev.txt
pip3 install --user -r only-dev.txt
cat pyproject.toml
cat only-dev.txt
```
```toml
[tool.poetry]
name = "pyMBE"
version = "0.8.0"
description = "Python-based Molecule Builder for ESPResSo"
authors = ["Your Name <you@example.com>"]
readme = "README.md"

[tool.poetry.dependencies]
python = "^3.8"
numpy = "^1.23"

[tool.poetry.group.dev]
optional = true

[tool.poetry.group.dev.dependencies]
pylint = "3.0.3"
coverage = "~7.4.4"
cmake = "^3.22.1"

[tool.poetry.group.docs]
optional = true

[tool.poetry.group.docs.dependencies]
pdoc = "^14.3"

[build-system]
requires = ["poetry-core"]
build-backend = "poetry.core.masonry.api"
```
```
astroid==3.0.3 ; python_version >= "3.8" and python_version < "4.0"
cmake==3.30.2 ; python_version >= "3.8" and python_version < "4.0"
colorama==0.4.6 ; python_version >= "3.8" and python_version < "4.0" and sys_platform == "win32"
coverage==7.4.4 ; python_version >= "3.8" and python_version < "4.0"
dill==0.3.8 ; python_version >= "3.8" and python_version < "4.0"
isort==5.13.2 ; python_version >= "3.8" and python_version < "4.0"
mccabe==0.7.0 ; python_version >= "3.8" and python_version < "4.0"
platformdirs==4.2.2 ; python_version >= "3.8" and python_version < "4.0"
pylint==3.0.3 ; python_version >= "3.8" and python_version < "4.0"
tomli==2.0.1 ; python_version >= "3.8" and python_version < "3.11"
tomlkit==0.13.2 ; python_version >= "3.8" and python_version < "4.0"
typing-extensions==4.12.2 ; python_version >= "3.8" and python_version < "3.11"
```

testsuite/henderson_hasselbalch_tests.py
jngrad: done

testsuite/henderson_hasselbalch_tests.py
jngrad: done

requirements.txt
jngrad: A comment line was added.

testsuite/henderson_hasselbalch_tests.py
```diff
@@ -16,87 +16,72 @@
...
+            ref_data_HH = np.loadtxt(f"{data_path}/HH.csv", delimiter=",")
+            np.testing.assert_allclose(Z_HH_1, ref_data_HH[0,:])
+            np.testing.assert_allclose(Z_HH_2, ref_data_HH[1,:])
+
+        with self.subTest(msg="Check Henderson-Hasselbalch equation + Donnan"):
+            HH_Donnan_dict = pmb.calculate_HH_Donnan(
+                    c_macro = {"peptide_1": pep1_concentration,
+                               "peptide_2": pep2_concentration},
+                    c_salt = c_salt,
+                    pH_list = pH_range)
```
jngrad, line 70: I took the liberty of replacing print statements by a corresponding `self.subTest()` context. This way, the message is only printed to the terminal when the assertion fails. Also, if the assertion fails, the test continues, so you can see how many equations are wrong. You no longer need to run the test multiple times with commented out sections to determine how many features are affected by a regression.

Note that `self.subTest()` is only useful when you cannot split assertions across multiple `def test_*()` functions, for example because all assertions rely on common data that is expensive to compute. If the data is cheap to generate, then it's preferable to use `def setUp()` and `def tearDown()` to set up the data before each test and clean it up at the end of each test. `tearDown()` will run even if the test failed.

testsuite/peptide_tests.py
```diff
@@ -16,87 +16,72 @@
...
+class Test(ut.TestCase):
 
-run_peptide_test(script_path=script_path,
-                    test_pH_values=test_pH_values,
-                    sequence=sequence,
-                    rtol=rtol,
-                    atol=atol)
+    def test_peptide(self):
+        with multiprocessing.Pool(processes=2) as pool:
+            results = dict(pool.map(kernel, tasks, chunksize=2))
```
jngrad, line 64: Some of the longer tests now leverage multiprocessing. We won't get better performance with more than 2 processes: for peptides, histatin-5 takes as long as the two short peptides, and for globular proteins, both take the same time. While this doesn't make much of a difference with `make tests -j4` (which is what we have in CI), with `make tests -j8` on laptops that have more than 4 cores, it's noticeably faster, because the CTest scheduler can do load balancing more efficiently. Until now, the two slowest tests took 20 min each, therefore CTest could not run any faster than 20 min. Now they take 12 min resp. 9 min each, which is in line with the third slowest test (15 min). 15 min is a huge improvement compared to the 50 min the old Makefile-based testsuite took.

If we later choose to add more proteins or peptide sequences, the specific order of strings in `tasks` and the value of `chunksize` need to be adapted to optimize load balancing between the two processes. One can increase the number of processes in the Python file, but then the corresponding variable `THREADS` in the CTest config file need to reflect that. I did not introduce multiprocessing in `weak_polyelectrolyte_dialysis_test.py` since we have only 1 molecule, and the pH values have inhomogeneous runtimes, such that with 2 processes, one of the process remains idle for 50 seconds waiting for the other process to complete the simulation at pH 9.

pm-blanco (approved):
Looks good to me, thank you for the detailed explanation @jngrad. I took a look into `poetry` and I agree with you: it is an interesting tool but not what we need at the moment. I do not think that it is worth for now to add another layer of complexity to our library. Thanks for the refactoring of the HH tests, the `ut.TestCases` sounds like a quite convenient framework. We can try to use it more for our other tests!

PR #90, opened 2024-09-03
Two-bead parametrization for peptides

mariusaarsten:
Parametrization of two-bead model for peptides

### DIFF
--- a/parameters/peptides/Aarsten2024.json
+++ b/parameters/peptides/Aarsten2024.json
@@ -1,0 +1,369 @@
+{
+    "data": {
+    "CA": {
+      "object_type": "particle",
+      "name": "CA",
+      "q": 0,
+      "sigma": {"value": 1, "units": "reduced_length"},
+      "offset": {"value": 0.152, "units": "nm"},
+      "epsilon": {"value": 1, "units": "reduced_energy"}
+    },
+    "G": {
+      "object_type": "particle",
+      "name": "G",
+      "q": 0,
+      "sigma": {"value": 1, "units": "reduced_length"},
+      "offset": {"value": 0.152, "units": "nm"},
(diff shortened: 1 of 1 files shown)


pm-blanco (changes requested):
@mariusaarsten Please add the metadata field, see `parameters/peptides/Blanco2021.json`.  Do you know if your master thesis have been already been published at the NTNU repository, we should add the publication details in there.

pm-blanco:
To be discussed with @kosovan: do you think that is a good idea to add this to the library? This is the first complete parametrization set for all aminoacids for the two bead model that we would have in pyMBE. We have validated it against the potentiometric titrations in @mariusaarsten thesis, which is going to be publicly available at the NTNU repository. However, the peptide sequences were not having all the aminoacids parametrized here so it is, of course, not ensured that it will provide reliable results for any arbitrary sequence. However, having the parametrization here has the added value that we can later use it for the testing the perspective new feature on dendrimers that will build on Marius' work.

mariusaarsten:
@pm-blanco The thesis should be published when it is graded, but the deadline for grading is in the beginning of  November, so still quite a while. No need to add the metadata field before it is graded and published?

pm-blanco:
@mariusaarsten I think that this PR can wait until your thesis is published. The file will not be used by any other part of pyMBE for now so it will not create any merge conflict while we keep developing the library.

pm-blanco:
This pull request has become eventually obsolete since we did not continue to make progress in the dendrimer feature. We can revisit the issue of adding this new parametrization set for peptides as part of the functional tests for peptide-conjugated dendrimers in the future, when someone actively works on its development. For now I am closing this PR to avoid cluttering our list of active PRs.

PR #93, opened 2024-09-12
General lattice builder

jngrad:
Description of changes:
- Add a lattice builder to design hydrogels with custom chain sequences
- Add an interactive visualizer to inspect the 3D scaffold
- Add a tutorial to show how a diamond hydrogel is built


jngrad:
This draft PR contains a scaffold that @1234somesh can build upon to create the final feature. Among other things, we need to add support for pyMBE types and residues, tesselated lattices in non-cubic boxes, particle and bond insertion in ESPResSo systems. The lattice builder works in lattice units, i.e. the diamond unit cells goes from (0, 0, 0) to (4, 4, 4), and nodes are labeled using square brackets. Be sure to double-check consistency of this notation with the one used in material sciences (e.g. the [Miller index](https://en.wikipedia.org/wiki/Miller_index)), or opt for a different notation.

Open-ended questions:
- support for chains between nodes that are normally not connected in a diamond lattice
   - now: allowed with `strict=False`
- support for chains that start and end on the same node
   - now: allowed with `strict=False` and symmetric sequences, otherwise the stereocenter would be undefined
- support for more than one chain between two nodes
   - would require a new data structure for `self.chains`
   - adding a chain between two nodes would no longer overwrite the old chain (e.g. the default-constructed one)
   - figure out an interface to iterate over chains between two nodes, and delete a specific one or all of them
- support for open chains
   - the special case where a chain between two nodes is deleted is trivial (use the `del` keyword)
   - the general case with two sequences would require a new data structure for `self.chains`
- support for chains between monomers of chains
   - would require a new data structure for `self.chains` and a lot more code
- support for lattices other than diamond
   - probably not too difficult
   - look into standardized data formats to describe the lattice nodes and connectivity
- support for more than one bond type
   - probably difficult

pm-blanco:
@jngrad thank you for an impresive and fast work to provide a template for the pyMBE builder of hydrogel networks! The structure looks pretty good to me and I think that is sensible to have this feature as a separate library, at least while we are developing it. 

Maybe a couple of suggestions for @1234somesh:
- Use `pmb.create_molecule` to create the chains between nodes. This way you will get the support for creating lateral chains already existing in pyMBE. You can control the starting position where the chain will be created using the optional argument `list_of_first_residue_positions`. You will need to add a new optional argument `backbone_vector` so you can also pass the direction vector in which the chain should be created, which should be pased inside that function to `create_residue`. I would make it so that if the user does not provide a `backbone_vector` then it is random, which will keep the current API.
- Use `pmb.create_particle` to create the nodes.
- Use `create_bond` to bond the nodes to the end of the chains. This actually solves the last bullet point by @jngrad as it gives support for both harmonic and FENE bonds (and we can add support for other kind of bonds by extending the features of that function).

1234somesh:
I have few questions how to proceed with the task. The goal of the task is to use ```set_node``` and ```set_chain``` defined in the lattice.py file to create the network and not create the network using, ```pmb.create_particle``` to create the nodes and ```pmb.create_molecule``` to create the chain. This gives the freedom for the user to set the desired orientation of the chains (through arguments in ```set_chain()```, set the desired type of nodes and chains, and as well as length of each chain. The user can visualize that using ```draw_lattice()``` function. @jngrad mentioned that, we need to pass espresso objects to the arguments  "residue" in ```set_node()``` and "sequence" in ```set_chain()```, currently they are strings/list of strings. 

I was thinking that, we can modify ```set_chain``` and ```set_node``` such that they use ```pmb.create_particle``` and ```pmb.create_molecule``` to create the nodes and chains. Because, first creating nodes and chains using pmb.*, and then using set_chain and set_nodes does not make sense as a user. To use these function, user also needs to provide , for example, positions of nodes, list of first particle in the chains etc. 

The question basically is, how do I use ```set_chain``` and ```set_nodes``` in a correct way?

1234somesh:
Notes from on-person discussion with @pm-blanco:

We need to split the logic in my current hydrogel.py script into two:
- `pmb.define_hydrogel`: Define the topology of the hydrogel:
    -  Link particle names with nodes
    -  Link residue_lists with hydrogel chains
    -  Import the DiamondLattice library to get the network topology

pm-blanco (approved):
looks good to me

PR #98, opened 2024-10-30
move_protein not boolian in globular_protein.py

Zitzeronion:
Currently the argument is parsed as a float and raises an error if set `True` when given as argument. I understand that `--move_protein 1` would do the trick, but wouldn't it be better if the the type is bool, considering that the default is `False`?


pm-blanco:
@Zitzeronion, @paobtorres  is currently working to solve this issue and another bug we found while revisiting the feature to enable the motion of globular proteins as rigid objects. To avoid a merge conflict with her current branch, I will prefer to close this PR and address this issue in her fork. Thank you for the report and the path!

PR #99, opened 2024-11-05
include backbone vector in create_molecule

1234somesh:
### Added
- New optional argument `backbone_vector` enabling to build molecules along an input vector using `pmb.create_molecule` and `pmb.create_pmb_object`


pyMBE.py
```diff
@@ -1847,7 +1848,7 @@ def find_bond_key(self, particle_name1, particle_name2, use_default_bond=False):
         bond_keys = [particle_name1 +'-'+ particle_name2, particle_name2 +'-'+ particle_name1 ]
         bond_defined=False
         for bond_key in bond_keys:
-            if bond_key in self.df.values:
+            if bond_key in self.df["name"].values:
```
pm-blanco, line 1851: @1234somesh this change looks like you found a small bug in another function. Did you find any unexpected behavior when using the previous implementation?

pm-blanco (changes requested):
Overall, it looks good to me. However, we are missing a unit test to check this new feature. I suggest adding such unit test to `testsuite/define_and_create_molecules_unit_tests.py`. This unit test should create a molecule using `pmb.create_molecule()` with an input `backbone_vector` and check that the backbone of the corresponding molecule created in espresso follows the same vector as the input  `backbone_vector`.

pyMBE.py
```diff
@@ -873,7 +873,7 @@ def create_counterions(self, object_name, cation_name, anion_name, espresso_syst
                 print(f'Ion type: {name} created number: {counterion_number[name]}')
         return counterion_number
         
-    def create_molecule(self, name, number_of_molecules, espresso_system, list_of_first_residue_positions=None, use_default_bond=False):
+    def create_molecule(self, name, number_of_molecules, espresso_system, list_of_first_residue_positions=None, backbone_vector=None, use_default_bond=False):
         """
         Creates `number_of_molecules` molecule of type `name` into `espresso_system` and bookkeeps them into `pmb.df`.
```
pm-blanco, line 878: The new input argument `backbone_vector` needs to be added to the docstring.

pyMBE.py
```diff
@@ -3213,6 +3214,10 @@ def setup_df (self):
                 '': object},
             'l0': {
                 '': float},
+            'node_map':{
+                '':object},
+            'chain_map':{
+                '':object}
```
pm-blanco, line 3220: @1234somesh I believe this lines have leaked from your work in the hydrogel feature. Could you please remove them from this branch and add them in the branch where you work in the feature? Otherwise the unit test `testsuite/read-write-df_test.py` crashes because one needs to adapt it to the new columns in `pmb.df`.

pm-blanco (changes requested):
The unit test looks good to me! Once you document the new optional argument `backbone_vector` in the list of args of the docstring of `create_molecule` and fix the issue with the additional columns in `pmb.df`,  I think that we can merge this PR.

pm-blanco (approved):
Looks good to me. Congrats on your first successful PR @1234somesh!

pm-blanco:
@kosovan @mariusaarsten here we have added the feature to create molecules along an input vector that you both needed for your projects.

PR #101, opened 2024-11-12
Add unit tests for globular proteins features

paobtorres:
### Added
- Unit tests for `pmb.create_protein`, `pmb.enable_motion_of_rigid_object`, `pmb.protein_sequence_parser`, `pmb.define_protein`, `pmb.read_protein_vtf_in_df` 

### Fixed
- Wrong parsing in `pmb.protein_sequence_parser` of input sequences provided as a list of aminoacids using the three letter code. 
- Wrong setup of the rigid object in `pmb.enable_motion_of_rigid_object`, leading to crashes in `samples/globular_protein.py` when enabling the protein motion.
- The argparse argument `--move_protein` of `samples/globular_protein.py` is now a boolean flag instead of taking arbitrary float values.


pm-blanco (approved):
@paobtorres I have reviewed the code and overall it looks good to me. I had to do a few changes because some parts were not fully covered:
- In the functional test for the feature to have a moving protein, small ions where also created. Therefore, the test would always pass even if the protein would not move because the small ions motion would cause the trajectory to be different. I fixed that by setting up an ideal system with no small ions for this test.
- The unit tests for globular proteins, I changed the test globular protein to 1f6s because in this way it also covers the case in which a metal ion is present in the PDB.
- Some corner cases where still missing in `protein_sequence_parser`
- I merged all unit tests for globular proteins in a single file `globular_protein_unit_tests`
Take a look to my changes and if you are OK then I think we are now ready to merge this PR

pm-blanco:
@Zitzeronion here we have fixed the issue you raised in #98

@jngrad once we merge this PR, we finally have a robust testing of the features of globular proteins so you can move forward with #13 when you have the time. The core features of the code are now covered, although we still need to keep working on improving the coverage until we cover all the corners of the code.

pm-blanco (approved):

PR #106, opened 2024-11-28
Lift restrictions on ESPResSo/NumPy/Pandas version requirements

jngrad:


README.md
```diff
@@ -129,24 +150,30 @@ jupyter kernelspec uninstall pymbe
...
-### Run the tutorial of pyMBE
+### Run tutorials
 
-You can run the interactive tutorial of pyMBE with the command:
+You can run the interactive tutorials with the following commands:
 
 ```sh
 source pymbe/bin/activate
-jupyter-lab tutorials/pyMBE_tutorial.ipynb
+jupyter-lab
```
jngrad, line 171: The old syntax would start the Jupyter server inside the `tutorials` folder and open `pyMBE_tutorial.ipynb`. While it appears to be convenient, the notebook was actually unable to display the diagrams, because relative paths of the form `../figs/image.png` are blocked by web servers for security reasons (otherwise a malicious actor would be able to traverse the entire file system of the machine the notebook is running on). Starting Jupyter in the top-level directory makes the `figs/` folder walkable by the web server.

README.md
```diff
@@ -94,9 +104,20 @@ deactivate
...
 
-We highlight that the module files need to be loaded before every activation
+Please note the module files need to be loaded before every activation
 of the virtual environment.
 
+NumPy 2 users need to override the version requirements as follows:
+
+```sh
+sed -i "/numpy/d" requirements.txt
+python3 -m pip install -r requirements.txt "numpy==2.1" "pandas>=2"
```
jngrad, line 114: This sed expression is unfortunate, but I couldn't think of a better alternative. There is a proposal to introduce a pip flag [`--ignore-conflicts`](https://pip.pypa.io/en/latest/ux-research-design/research-results/override-conflicting-dependencies/#recommendations) that would allow users to override version requirements from `requirements.txt` by specifying new requirements as command line arguments. A draft implementation was proposed, but it's currently stalled due to the complexity of gracefully handling edge cases where no suitable version of a package can be determined due to version requirements of indirect dependencies not being satisfied. Providing two different requirement files goes against pip best practices. Forcing users to migrate to NumPy 2 is also not an option, since cluster users must use NumPy 1 when loading modules from EasyBuild or EESSI.

README.md
```diff
@@ -94,9 +104,20 @@ deactivate
...
+```
+
+Please note that for NumPy 2 to be supported, ESPResSo needs to be built
+against NumPy 2 too, otherwise it will complain about API version mismatch.
+Be aware the testsuite of ESPResSo 4.2 is known to be flaky with NumPy 2.
```
jngrad, line 119: I wanted to better separate all 3 alternatives (NumPy 1 users, cluster users, NumPy 2 users), using something similar to [sphinx tabs ](https://sphinx-tabs.readthedocs.io/en/latest/) like on the official Cython website ([example](https://cython.readthedocs.io/en/latest/src/tutorial/cython_tutorial.html#primes)), but the GitHub-flavored Markdown syntax doesn't support it. I've seen feature requests on GitHub support pages where people pointed out [R Markdown](https://bookdown.org/yihui/rmarkdown-cookbook/html-tabs.html) could do that, but it's unclear whether this is being worked on. Wrapping the alternatives into [collapsible sections](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/organizing-information-with-collapsed-sections) doesn't feel right to me, because this feature is typically used to hide "non-critical" information (long scripts or error logs) in bug reports to make them more readable.

README.md
jngrad: Maybe we could do the following:
```sh
# for NumPy 1 users
python3 -m pip install -r requirements.txt "numpy<2.0" "pandas<2.0"
```
```sh
# for NumPy 2 users
python3 -m pip install -r requirements.txt "numpy>=2.1" "pandas>=2.0"
```

This way both user communities have a similar-looking pip install command, and we can remove the `<2` restriction from `requirements.txt`. This would also better reflect how pyMBE supports both NumPy releases.

pm-blanco: @jngrad I think that I like this last suggestion more. Using sed can be a bit "blackboxy" for people not so familiar with bash commands, which could potentially confuse some of our users.

jngrad: Done!

README.md
jngrad: This was fixed by merging the NumPy 2 paragraph into the first paragraph.

samples/Beyer2024/globular_protein.py
```diff
@@ -87,6 +87,8 @@
 inputs={"pH": args.pH,
         "pdb": args.pdb}
 
+steps_name = f"{espressomd.version.friendly() == '4.2' and 'reaction_' or ''}steps"
```
pm-blanco, line 90: While this is a very compact and elegant way of solving the change in the API of espresso, it might be a bit cryptic for people missing the context. What about expanding the code a bit and adding a clarifying comment?
```python
# To keep backwards compatibility due to an API change from espresso 4.2 to espresso 4.3
if espressomd.version.friendly() == '4.2':
       keyword_mc_reaction="reaction_steps"
else:
       keyword_mc_reaction="reaction"
```
If you agree then the function should also be changed in the other sample scripts for consistency.

samples/peptide_mixture_grxmc_ideal.py
```diff
@@ -60,13 +61,17 @@
...
 # The trajectories of the simulations will be stored using espresso built-up functions in separed files in the folder 'frames'
 Path("./frames").mkdir(parents=True, 
                        exist_ok=True)
 
+steps_name = f"{espressomd.version.friendly() == '4.2' and 'reaction_' or ''}steps"
+
+def get_number_of_particles(espresso_system, ptype):
+    args = (ptype,) if espressomd.version.friendly() == "4.2" else ()
+    kwargs = {} if espressomd.version.friendly() == "4.2" else {"type": ptype}
+    return espresso_system.number_of_particles(*args, **kwargs)
```
pm-blanco, line 73: In the same spirit as before maybe expanding the code a bit for better readability?
```python
def get_number_of_particles(espresso_system, ptype):
    """
    Gets the number of particles of a given type `ptype` from espresso, handling changes recent changes in the API between the version 4.2 and 4.3 of espresso.
    """
    if espressomd.version.friendly() == "4.2":
                 args = (ptype,)
                 kwargs = {}
    else:
                 args = ()
                 kwargs = {"type": ptype}
    return espresso_system.number_of_particles(*args, **kwargs)
```
If you agree then the function should also be changed in the other sample scripts for consistency.

pm-blanco (commented):
@jngrad thank you for taking care of updating our dependencies and solving some typos in the way! I checked and everything works with espresso v4.3 and v4.2. This is an important step forward enabling to combine pyMBE with other tools as MDAnalysis, opening the door of our goal to make our software more interoperable. I only have some minor comments:

samples/Beyer2024/globular_protein.py
jngrad: I considered it, but was faced with the issue of code coverage decreasing, since we don't run CI with ESPResSo 4.3-dev yet. If you're fine with that, I will apply the changes you propose.

I was also wondering whether to implement free functions like `setup_electrostatic_interactions()` that take the system as argument and the solver or particle type, and let this function go through the if/else for us. This way we reduce code duplication and make pyMBE compatible with multiple ESPResSo releases. The downside is that we end up creating a new ESPResSo interface that pyMBE devs now have to maintain.

pm-blanco: I was just reflecting about the same actually.  To me it boils down to the same dilemma than with the rest of "convenience" functions  in `lib/handy_functions.py`: we reduce code repetition at the cost of creating an extra layer on top of espresso that is arguably not needed and sometimes more limited that the native functions. I guess that If we give the added value of helping our users with backwards compatibility with earlier versions of ESPResSo, it could make some sense to keep it. I have not yet came to an absolute decision on this topic, but I suggest that for consistency with the current version we go for the strategy of having those functions in `libs/handy_functions.py`. In the long run, if we decide to keep that library we then need to integrate it into our CI pipeline and maintain it properly.

jngrad: Done.

pm-blanco (approved):
Looks good to me. Now the handy_tools library is exposed to Codecov, which reduces the code coverage. It makes me sad to see the code coverage go down but at the same time the estimate of the code coverage is now more fair, since all libraries are included, improving our transparency. Let's keep working on improving the coverage of the library with more CI test to bring the coverage of the library back up!

PR #109, opened 2025-01-17
Improve visibility of the docs and license

pm-blanco:
Adds:
- Badges for the documentation and license
- An explicit link to our docs in `README.md`

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,7 @@
 
 ![GitHub Actions](https://github.com/pyMBE-dev/pyMBE/actions/workflows/testsuite.yml/badge.svg)
 [![codecov](https://codecov.io/gh/pyMBE-dev/pyMBE/branch/main/graph/badge.svg)](https://codecov.io/gh/pyMBE-dev/pyMBE)
+![Static Badge](https://img.shields.io/badge/available-a?style=for-the-badge&label=docs&labelColor=%23FAE442&color=%23C2F0BD&link=%20pymbe-dev.github.io%2FpyMBE%2FpyMBE.html%20)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,10 +8,14 @@
 
 ![GitHub Actions](https://github.com/pyMBE-dev/pyMBE/actions/workflows/testsuite.yml/badge.svg)
 [![codecov](https://codecov.io/gh/pyMBE-dev/pyMBE/branch/main/graph/badge.svg)](https://codecov.io/gh/pyMBE-dev/pyMBE)
+[![docs](https://badgen.net/static/docs/up-to-date/yellow/)](pymbe-dev.github.io/pyMBE/pyMBE.html)
```
jngrad, line 11: From a user experience perspective, the color yellow is often associated with low code coverage, low code quality, partially interrupted services and pending changes. In particular, the competing service readthedocs.io uses a very similar badge with yellow when the software documentation is left in an indeterminate state (user manual page on [Status badges](https://docs.readthedocs.io/en/stable/badges.html)). Not to mention the difficulty of reading white text on a yellow background.

I would personally vote for `blue`, which would look like this: [docs](pymbe-dev.github.io/pyMBE/pyMBE.html)

pm-blanco: what about having it on `purple` [docs](pymbe-dev.github.io/pyMBE/pyMBE.html) ?

Just because we are already using blue for the license

pm-blanco: Fine, let's have it blue for consistency between static banners. Thanks for sharing that post, that made me laugh!

jngrad (approved):

PR #111, opened 2025-02-10
remove obsolete functions from `handy_functions`

pm-blanco:
### Removed
- `handy_functions.create_random_seed` no longer needed because now instances of pyMBE take the random seed as input 
- `handy_functions.visualize_espresso_system` because it was not used anywhere in the library 
- `handy_functions.do_snapshot_espresso_system` moved to the tutorial because it was a function specific for it


tutorials/pyMBE_tutorial.ipynb
```diff
@@ -231,6 +227,28 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "# Only necesary to produce the pictures used in this tutorial\n",
+    "from PIL import Image\n",
+    "def do_snapshot_espresso_system(espresso_system, filename):\n",
+    "    \"\"\"\n",
+    "    Uses espresso visualizator for creating a snapshot of the current state of the espresso_system\n",
```
jngrad, line 234:
```suggestion
    "# Only necessary to produce the pictures used in this tutorial\n",
    "from PIL import Image\n",
    "def do_snapshot_espresso_system(espresso_system, filename):\n",
    "    \"\"\"\n",
    "    Uses espresso visualizer for creating a snapshot of the current state of the espresso_system\n",
```

tutorials/solution_tutorial.ipynb
```diff
@@ -28,24 +28,32 @@
...
+     "text": [
+      "Current set of reduced units:\n",
+      "0.35 nanometer = 1 reduced_length\n",
+      "4.1164e-21 joule = 1 reduced_energy\n",
+      "1.6022e-19 coulomb = 1 reduced_charge\n",
+      "Temperature: 298.15 kelvin\n",
+      "The side of the simulation box is  7.5 nanometer = 21.428571428571427 reduced_length\n"
+     ]
+    }
+   ],
```
jngrad, line 46: don't forget to clear output in the tutorial

tutorials/solution_tutorial.ipynb
```diff
@@ -197,11 +205,33 @@
...
+   "execution_count": 6,
    "metadata": {},
    "outputs": [],
    "source": [
-    "picture_name = 'alternating_system.png'\n",
+    "# Only necesary to produce the pictures used in this tutorial\n",
+    "from PIL import Image\n",
+    "def do_snapshot_espresso_system(espresso_system, filename):\n",
+    "    \"\"\"\n",
+    "    Uses espresso visualizator for creating a snapshot of the current state of the espresso_system\n",
```
jngrad, line 216:
```suggestion
    "# Only necessary to produce the pictures used in this tutorial\n",
    "from PIL import Image\n",
    "def do_snapshot_espresso_system(espresso_system, filename):\n",
    "    \"\"\"\n",
    "    Uses espresso visualizer for creating a snapshot of the current state of the espresso_system\n",
```

tutorials/pyMBE_tutorial.ipynb
```diff
@@ -231,6 +227,28 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "# Only necesary to produce the pictures used in this tutorial\n",
+    "from PIL import Image\n",
+    "def do_snapshot_espresso_system(espresso_system, filename):\n",
```
davidbbeyer, line 232: Just a nomenclature thing, but I would call this function `create_snapshot_espresso_system`

tutorials/solution_tutorial.ipynb
davidbbeyer: Related to this: currently there are separate files for the tutorial and solution, so there is always the possibility of introducing incompatibilities between these two files. Is there some way to avoid this, e.g. a single file with initially hidden solutions as we do it for the ESPResSo tutorials?

davidbbeyer (commented):
I agree with the comments raised by @jngrad. Another thing I noticed when looking at `handy_functions.py` again: the `print()` statement in line 64 seems superfluous. Could you check and remove it?

tutorials/solution_tutorial.ipynb
pm-blanco: @davidbbeyer what incompatibilities are you specifically referring to? Do you mean inconsistencies between naming of functions and variables or some other more serious issue?

I do not completely oppose to the idea of merging the solutions into the tutorial, although it does come with some drawbacks:
- The feature with the hidden solutions is an extension for jupyter notebooks. I am not sure if we can enforce that such feature is active when using the pyMBE kernel @jngrad? Would that potentially create some issues for our users?
- My experience when I see people doing the ESPResSo  is that having the solution "one click away" becomes too tempting and they tend to simply read it before trying to solve the exercise themselves.  Then again, having the solution "one file away" is not really preventing that much more...

davidbbeyer: What is mean is that basically the content (e.g. function names) could diverge between the two files, since we are not explicitly comparing them or deriving them from a single file.

pm-blanco: I do not see any clear solution to this issue. Of course, it would be better to avoid inconsistencies between the tutorial and the solution but I am not convinced by the alternatives either. I think that this discussion is out of the scope of this PR, so I propose that we discuss it in our next monthly meeting and we move forward with this PR unless there is something else to be fixed here.

davidbbeyer (approved):

PR #112, opened 2025-02-10
code-cov pmb.setup_lj_interaction

1234somesh:
Partial fix to #58 and #115.

### Added
- Helper method to delete entries from `pmb.df` 

### Fixed
- Warning handling and coverage in `setup_lj_interaction`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Switched to Ctest for testing, allowing to run the tests on paralel (#87)
 
 ### Added
+- Helper method to delete entries from `pmb.df` (#112)
 - Code of conduct of our community `CODE_OF_CONDUCT.md`, adhering to the Contributor Covenant v2.1 (#104) 
 - New optional argument `backbone_vector` enabling to build molecules along an input vector using `pmb.create_molecule` and `pmb.create_pmb_object` (#99)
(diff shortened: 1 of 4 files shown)


pm-blanco (changes requested):
Thank you for your work here @1234somesh!

There are a few minor points to consider before merging this to main:
- I think that `delete_particle_entry` actually is a more general function than its name suggests: it will actually delete _any_  entry with the input name, regardless if that name is a particle, residue, molecule... 
- There is now somewhat of a code redundancy between `delete_particle_entry` and `destroy_pmb_object_in_system`. The issue is that `destroy_pmb_object_in_system` is now both deleting molecules from the espresso system and deleting entries from the pyMBE, which makes pyMBE not very modular.  

I suggest the following solution:
- Rename `delete_particle_entry` to `delete_entries_in_df`  and `particle_name` to `entry_name`
- Adapt the docs of `delete_entries_in_df` to explain that it will delete all entries in `pmb.df` with  name =  `entry_name`
- Do a CI test that checks that `delete_entries_in_df` works as intended. 
- Update the years of license in `testsuite/lj_tests.py` to include 2025 (date of last modification).

I would leave the refactoring of `destroy_pmb_object_in_system` to a separate PR because it requires of careful refactoring that implies more work than simply making it call your function.

pm-blanco (approved):
Looks good to me. Good job with the tests @1234somesh!

pm-blanco (approved):

PR #113, opened 2025-02-14
Add unit tests for bonds

jngrad:
Description of changes:
- handle all edge cases in bond creation/search/serialization
- replace unsafe `ast.literal_eval()` by equivalent `json.loads()`
- improve code coverage (partial fix for #58)

### DIFF
--- a/pyMBE.py
+++ b/pyMBE.py
@@ -18,7 +18,6 @@
 
 import re
 import sys
-import ast
 import json
 import pint
 import numpy as np
@@ -649,7 +648,7 @@ class pymbe_library():
             if df[column_name].isnull().all():
                 df[column_name] = df[column_name].astype(object)
             else:
-                df[column_name] = df[column_name].apply(lambda x: ast.literal_eval(str(x)) if pd.notnull(x) else x)
+                df[column_name] = df[column_name].apply(lambda x: json.loads(x) if pd.notnull(x) else x)
 
         for column_name in columns_with_units:
             df[column_name] = df[column_name].apply(lambda x: self.create_variable_with_units(x) if pd.notnull(x) else x)
@@ -659,7 +658,6 @@ class pymbe_library():
         return df
     
     def convert_str_to_bond_object (self, bond_str):
-        
         """
         Convert a row read as a `str` to the corresponding bond object. There are two supported bonds: HarmonicBond and FeneBond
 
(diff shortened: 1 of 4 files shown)


pm-blanco (changes requested):
In `search_bond`:
- there is an uncovered line in 2549 that only triggers when the switch `hard_check` is active:
```python
                sys.exit(1) 
```
In `get_bond_length`:
-  Same issue as before in line 2003
 
Other than that, it looks very good to me. Thank you for taking the initiative to take care of this @jngrad. 

Could you also update the Changelog?

jngrad:
> there is an uncovered line in 2549 that only triggers when the switch hard_check is active

This is meant to be that way. If the original authors intended these errors to be recoverable, they would have opted to raise an exception instead of triggering a fatal error.

I'm planning to open a ticket next week to discuss this topic and propose a more sustainable alternative, because it's not a good design. The Python session crashes without producing a traceback, making it impossible for users to understand what caused the fatal error. It's also untestable: you can try catch a `SystemError` to intercept `sys.exit(1)`, but the pmb object is now in an indeterminate state and cannot be used safely. In fact, the `tearDown()` method might call a pmb method to clean up the data.frame, and thus trigger undefined behavior.

As part of this discussion, I'll also touch upon the print statements, which print warnings to stdout instead of stderr, making it more difficult to investigate issues in a slurm job where stdout and stderr and redirected to different files. In addition, print statements don't show the filename and line number to help you diagnose the source of the error, something that `logging.warn()` does. The advantage of `logging` is that you can set a severity level for each message, and users can decide globally which severity level should raise an exception. This way, they don't need to pass `hard_check` to methods. These logs can also be captured in unit tests, which is more convenient than using a contextlib to temporarily redirect stdout to a stream (not to mention the surprise if you capture more than 1 print statement in the same stream).

I'll look into producing a MWE next week, when I have more time to experiment with `unittest` and `logging`.

pm-blanco:
> > there is an uncovered line in 2549 that only triggers when the switch hard_check is active
> 
> This is meant to be that way. If the original authors intended these errors to be recoverable, they would have opted to raise an exception instead of triggering a fatal error.

I can tell you that the original author simply did not think too much about it when wrote those lines ;) I agree that we should replace this logic with a better code design, that is also why I raised this issue. What you propose sounds like a much more cleaner solution, I like it. Let's then leave those lines uncovered by now and add coverage once we refactor the error handling.

pm-blanco (approved):

PR #120, opened 2025-03-19
bugfix: support `backbone_vector` as a list

pm-blanco:
Small fix to solve a bug reported by @nejedlama.  

### Fixed
- Occasional crashes when the user provided `backbone_vector` as list instead than a `ndarray` in `pmb.create_molecule()`


pinedaps (commented):
Looks OK. Nevertheless, It seems a bit more straightforward to me to put it as the else option of the already existing "if backbone_vector is None:" in the line 908 in 
 the case of pmb.create_molecule() and in the existing else for pmb.create_residue() in line 1193.

pinedaps (approved):
Looks Great!

PR #125, opened 2025-05-14
Increase code coverage for various functions

davidbbeyer:
I have increased the code coverage of the following unit tests:

- `calculate_HH` and `calculate_HH_Donnan` are now tested also in the case when no `pH_list` is provided
- `determine_reservoir_concentrations` is now also tested for the interacting case, before it was only tested for an ideal system. The test now also covers the while statement, which was previously untested. Because the calculation takes longer in the interacting case, we only consider a single reservoir composition in the unit test.
- `generate_trial_perpendicular_vector` is now explicitly tested for a ValueError in the case when the user provides a zero vector.


pm-blanco (approved):
looks good to me. Thank you for your contribution to the CI @davidbbeyer

PR #131, opened 2025-05-28
Modernize

jngrad:
Reduce code duplication and add support for the development version of ESPResSo.


jngrad:
@pm-blanco This is the preparation work for restructuring pyMBE as a package. I'm splitting the work to make the refactoring PR easier to review. The small changes in this PR are often found on the lines above and below the lines containing pyMBE paths, and they would be incredibly distracting if mixed in with the changes of the upcoming PR.

pm-blanco (approved):
Looks good to me, thank you for the errata correction and code modernization!

PR #132, opened 2025-05-30
Restructure pyMBE as a package

### LINKED ISSUE #13: Convert pyMBE into a package

jngrad:
Fixes #13

Description of changes:
* parameter files and the `lib/` folder now belong to the pyMBE package
* all other files (samples, tutorials, maintainer scripts, paper scripts, etc.) still belong to the pyMBE project, but they don't get deployed when installing the pyMBE package
* function `pmb.get_resource()` was removed


jngrad:
Note: pdoc automatically creates an empty landing page for package modules. Comment lines were added in the readme file to populate the landing page with something useful. Here is a preview of the new documentation:
https://jngrad.github.io/pyMBE/pyMBE.html

You can test packaging locally by creating a new file `setup.py` at the top-level of your git clone and populating it with:
```python
from setuptools import find_packages, setup

with open("README.md", "r") as f:
    long_description = f.read()

with open("requirements.txt") as f:
    required = f.read().splitlines()

setup(
    name="pyMBE",
    version="0.8.0",
    author="David Beyer, Pablo M. Blanco, et al.",
    author_email="pablb@ntnu.no",
    description="Toolkit to facilitate building up molecules with complex architectures in the molecular dynamics software ESPResSo.",
    long_description=long_description,
    long_description_content_type="text/markdown",
    packages=find_packages(),
    python_requires=">=3.10",
    install_requires=required,
)
```
Then install pyMBE in a fresh virtual environment with:
```sh
python3 -m venv venv-local
source venv-local/bin/activate
python3 maintainer/configure_venv.py --espresso_path=/home/jgrad/espresso/build # adapt path
pip install .  # install pyMBE and its dependencies automatically
cd testsuite   # important: leave the current folder! otherwise Python imports the local pyMBE folder
python3 bond_tests.py
```

Please note `setup.py` is deprecated in favor of toml files. For more details, see [this chapter](https://pip.pypa.io/en/stable/reference/build-system/setup-py/) in the pip documentation. The example above is only meant as a sanity check. Packaging pyMBE on PyPI is not within the scope of this PR.

PR #133, opened 2025-06-25
Run tests against ESPResSo 5.0-dev

jngrad:


jngrad:
Looks like the testsuite is slower than expected. I think I forgot to set the number of OpenMP threads. With `pypresso`, the threads are set automatically for the user, but in pyMBE we don't use `pypresso` in favor of `python3`. Let me have a look.

jngrad:
Here is the `samples` workflow dispatch: run [15879324788](https://github.com/jngrad/pyMBE/actions/runs/15879324788)

Note for maintainers: we now use a [workflow matrix](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow) to run the same testsuite on multiple configurations. However, only one configuration is allowed to upload the documentation files as an artifact. I'm using ESPResSo 4.2.2 for that, since it's the last stable release. Once we drop support for ESPResSo 4.2.2, we will test against 5.0.0 and 5.1-dev, and we'll need to update which version of ESPResSo uploads the documentation files as an artifact.

I've also updated the virtual environment configuration to inject OpenMP-specific environment variables, which are mandatory when ESPResSo 5.0 is built with `-D ESPRESSO_BUILD_WITH_SHARED_MEMORY_PARALLELISM=ON`. I pinned the number of cores to 4, since it's what is available on the [GitHub specifications](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories); the `nproc` command now returns 1 instead of 4 because it gets overridden by environment variables `OMP_NUM_THREADS` (see `nproc` [documentation](https://www.gnu.org/software/coreutils/manual/html_node/nproc-invocation.html) on that topic).

pm-blanco (approved):
Looks good to me, thanks for your work!

PR #134, opened 2025-06-26
General maintenance tasks

jngrad:
Improve samples workflow runtime, add conda support, and fix a few bugs.


maintainer/configure_venv.py
```diff
@@ -35,13 +43,31 @@ def make_pth(name, path):
         f.write(os.path.realpath(path))
 
 
-parser = argparse.ArgumentParser(description="Configure pyMBE and ESPResSo module paths")
-parser.add_argument("--espresso_path", type=str, required=not espressomd_found,
-                    help="Path to the ESPResSo build folder")
-args = parser.parse_args()
+class PyVirtualEnv(enum.StrEnum):
```
pm-blanco, line 46: StrEnum is only supported in Python 3.11 and newer and currently we support Python 3.10+. Due to this change I could not setup pyMBE neither with microconda or venv (using Python 3.10.12) .

The following fix solved the issue for me:
```python
import os
import stat
import enum
import argparse
import sysconfig

#  Backport for StrEnum (if using Python < 3.11)
try:
    from enum import StrEnum
except ImportError:
    class StrEnum(str, enum.Enum):
        pass

#  Replace usage of enum.StrEnum with just StrEnum below
class PyVirtualEnv(StrEnum):
    venv = enum.auto()
    conda = enum.auto()

```

pm-blanco (changes requested):
The issue with the compatibility with Python 10 needs to be fixed in this PR. The patch for relax_espresso_system for now solves the issue of the exponentially increasing value of gamma, but we should change that logic in a dedicated PR. I will take care of this last point immediately after we merge this PR.  The rest looks OK to me.

pm-blanco (changes requested):
Seems that I approved the PR too quickly... the EESSI setup seems to be failing now.

maintainer/configure_venv.py
jngrad: Done. But please keep in mind that Python 3.10 was dropped from the NumPy ecosystem in April 2025, in accordance with the NEP 29 [drop schedule](https://numpy.org/neps/nep-0029-deprecation_policy.html#drop-schedule). NumPy 1.x will no longer be supported as of September 2025 and was already dropped from Ubuntu 25.04.

Also, the proposed change will make the error message cryptic: "This script should be run in a Python virtual environment, but none was detected (tried 1, 2)" instead of "This script should be run in a Python virtual environment, but none was detected (tried venv, conda)".

jngrad:
The GitHub Actions are failing due to https://github.com/actions/runner/issues/3720 (workflow job [44919627814](https://github.com/pyMBE-dev/pyMBE/actions/runs/15924829190/job/44919627814)). It's happening at random: the workflow works in my fork. Let me restart the workflow here.

maintainer/configure_venv.py
pm-blanco: I am afraid that we cannot expect our users to be completely up to date with the NEP 29 drop schedule as some may want to use pyMBE for projects that rely in software that is not up to date. 
We can discuss when would be appropriate to drop support in Python 3.10 in our next community meeting, but for the sake of this PR we should not change our requirements.

Regarding the error message, could we not find a solution to make it more explicit?

jngrad: I chose `StrEnum` precisely because a string-based solution like `if venv_family == "venvv"` is sensitive to typos, not to mention subtleties like `"conda"` vs `"miniconda"`. I'm not really willing to put more effort into re-inventing the wheel: Python 3.11 fixed that problem a while ago, and conda users can create an environment with a newer Python version trivially. In fact, by following the instructions in the pyMBE readme file, conda automatically forced the installation of a newer Python version for me.

pm-blanco: What would be your proposed solution then? We cannot rely on `StrEnum` and keep supporting Python 3.10. At the same time, I do not think that this PR is the right moment to change our dependencies. I agree that a solution beyond simply using strings is not worth the developer time since we will drop support for Python 3.10 in the future but I understand that they are not really a reliable solution. Seems that the cryptic message is the lesser evil for now, but it does not feel like a very satisfying solution...

pm-blanco (approved):
I tested the new implementation with Python 3.10 both setting up the virtual environment with venv and with miniconda3 and it worked for me.

PR #141, opened 2025-10-17
Remove legacy virtual environment overrides

jngrad:
By default, OpenMP applications use all CPU cores unless one exports environment variable `OMP_NUM_THREADS=1`, but doing so changes the behavior of Linux command `nproc` to return that variable value instead of the actual number of CPU cores. This can be confusing to users who aren't familiar with OpenMP parallelization. Most clusters will build ESPResSo 5.0 with OpenMP enabled for performance reasons. The latest version of ESPResSo automatically sets the number of OpenMP threads to 1 a few milliseconds after `import espressomd` is executed, if that environment variable wasn't set by the user. We therefore no longer have to alter environment variables in the user's Python environment.

### DIFF
--- a/.github/workflows/samples.yml
+++ b/.github/workflows/samples.yml
@@ -21,7 +21,7 @@ jobs:
           - version: "4.2.2"
             eessi: ESPResSo/4.2.2-foss-2023a
           - version: "5.0-dev"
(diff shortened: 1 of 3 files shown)


pm-blanco (approved):
Thank you for the maintenance work @jngrad. This is a very welcomed change!

PR #147, opened 2026-02-10
Replace the pyMBE  Pandas dataframe for a canonical pyMBE database

pm-blanco:
Fixes #146, #123 and #18 : 

## Added
- Introduced a canonical pyMBE database backend replacing the previous monolithic Pandas DataFrame storage approach. This lays the foundation for more robust, extensible, and normalized data handling across pyMBE.
- Added support to define reaction templates  in the pyMBE database.
- Utility functions to cast information about templates and instances in the pyMBE database into pandas dataframe `pmb.get_templates_df`, `pmb.get_instances_df` and `pmb.get_reactions_df`.
- Utility functions to load and save the new database via the pyMBE API, `pmb.save_database` and `pmb.load_database`.
- Added functions to define particle states:  `pmb.define_particle_states`  and `pmb.define_monoprototic_particle_states`.
- Added utility functions in `lib/handy_functions` to define residue and particle templates for aminoacids en peptides .and residues: `define_protein_AA_particles`, `define_protein_AA_residues` and `define_peptide_AA_residues`.

## Changed
- Create methods (`create_particle`, `create_residue`, `create_molecule`, `create_protein`, `create_hydrogel`) now raise a ValueError if no template is found for an input `name` instead than a warning.
- Refactored core modules to use the new database schema based on templates and instances  for particles, residues, molecules, hydrogels, proteins and peptides.
- Particle states now are independent templates and are now disentangled from particle templates. 
- Pka values are now stored as part of chemical reactions and no longer an attribute of particle templates.
- Amino acid residue templates are no longer defined internally in `define_peptide` and `define_protein`. Those definitions are now exposed to the user.
- Molecule templates now need to be defined to be used as templates for hydrogel chains in hydrogels.

## Fixed
- Wrong handling of units in `get_radius_map` when the `dimensionless` argument was triggered.
- Utility methods `get_particle_id_map`, `calculate_HH`, `calculate_net_charge`,  `center_object_in_simulation_box` now support all template types in pyMBE, including hydrogels. Some of these methods have been renamed to expose directly in the API this change in behavior.

## Removed
- Method `add_bonds_to_espresso` has been removed from the API. pyMBE now adds bonds internally to ESPResSo when molecule instances are created into ESPResSo.
- Tutorial `lattice_builder.ipynb` has been removed because its content is redundant with sample script `build_hydrogel.py`.

## Documentation
- Cleaned and standardized documentation.

## Tests
- Modernized molecule tests and LJ tests.

Issue #143, opened 2025-10-21
Add support for all bonded potentials


pm-blanco:
Currently only harmonic and FENE  bond interactions are supported in pyMBE. That limits our capabilities as a tool to setup advanced force fields in ESPResSo. Tasks to be done to make progress in this project:

- [ ] Implement `define_angular_potential` and `define_dihedral_potential` functions in analogy to our current setup for bonds, done with `define_bond`.
- [ ] Implement bookkeeping tool (map) of the connectivity between pairs, triplets and quadriplets of particles within molecules either using the pyMBE dataframe (one would need to create two more columns for particle ids) or with a dedicated data structure.
- [ ] When creating a molecule, pyMBE should check for the conectivity of the particle and add bending and dihedral potentials (if defined).

pm-blanco:
Draft of  `define_angular_potential` , after online discussion with @1234somesh 

- We add support for "Harmonic angle", "Cosine angle" and "Harmonic cosine" potentials, given by the `angle_potential_type`
- particle_triplets = [ [name0, name1, name2 ],  ], where we hard-code that the name in the center of the inner list is always the central particle. This should be clearly documented.
- We also define a `create_angular_potential_in_espresso(self, angle_potential_type, angle_parameters)` that checks the parameters and creates the espresso angular potential object (cf. `create_bond_in_espresso`)
- the angle_key of a particle triplet ["A", "B", "C"] should be "A-B-C"

```python
 def define_angular_potential(self, angle_potential_type, angle_parameters, particle_triplets):
        # Add docs here
        angular_potential_object=self.create_angular_potential_in_espresso(angle_potential_type, angle_parameters)
        for side_particle0, central_particle, side_particle1 in particle_triplets :
            index = len(self.df)
            # Adapt this to the label for angular potential A-B-C
            for label in [f'{particle_name1}-{particle_name2}', f'{particle_name2}-{particle_name1}']:
                self._check_if_multiple_pmb_types_for_name(name=label, pmb_type_to_be_defined="angle_potential")
            name=f'{particle_name1}-{particle_name2}'
            self._check_if_multiple_pmb_types_for_name(name=name, pmb_type_to_be_defined="angle_potential")    
            self.df.at [index,'name']= name
            self.df.at [index,'bond_object'] = bond_object      
            self.add_value_to_df(index=index,
                                    key=('pmb_type',''),
                                    new_value='angle_potential')
            self.add_value_to_df(index=index,
[...]

PR #150, opened 2026-04-08
Add support for angle potentials 

1234somesh:
Partially solves #143 
## Added
- Support to setup angular potentials with pyMBE. All flexible pyMBE templates now support angular potentials: hydrogels, molecules, peptides and residues (including residues with nested residues).
- Sample scripts and tests for the new functionality.
- Template and instance `Angle` to store information about angular potentials in the pyMBE database.
- Functions `define_angle` and ``define_default_angle` to define a templates of  angular potentials in pyMBE.
- Function `create_angle` to create instances of angular potential in pyMBE
- Function `get_angle_template` to retrieve a template of an angle potential in the pyMBE database.

## Fixed
- Regression introduced in #147, wrong handling of hydrogel chains with reverse order of residues.


testsuite/hydrogel_builder_with_angles.py
```diff
@@ -0,0 +1,186 @@
+#
+# Copyright (C) 2024-2026 pyMBE-dev team
```
pm-blanco: The copyright of the new files should only have the current year (2026)

samples/build_hydrogel_angle.py
```diff
@@ -0,0 +1,161 @@
+#
```
pm-blanco: Please add the copyright notice

samples/build_hydrogel_angle.py
```diff
@@ -0,0 +1,161 @@
...
+# This script builds a simple hydrogel of:
+#   - N: crosslinker/node particles
+#   - C: residue/chain particles
+#
+# By default, it defines only C-C-C and generates only chain-internal angles.
+# If run with `--include`, it also defines the crosslinker-adjacent
+# angle templates N-C-C and C-N-C.
+#
+# Note:
+#   - The canonical angle name for N-C-C is C-C-N, because pyMBE sorts the
```
pm-blanco: I find the example a bit confusing, when I first read it I thought that all three particles were being sorted instead of only the two sides because there are two "C" labels in the example. Perhaps it could be clearer using three different labels for the example. For example, for `particle_triplets = ["C", "D", "A"]' the canonical name for the angle template in the pyMBE database is `name = "A-D-C"`

samples/build_hydrogel_angle.py
pm-blanco: `samples/build_hydrogel_angle.py` was extremely helpful to validate the correct implementation of the bending potentials in the hydrogel case. However, it feels more like a "test" script than really a sample script (something a student or a new user would take to try out pyMBE). I suggest to merge this script to the already existing hydrogel sample `build_hydrogel.py`  and add an optional arguments to  control if one wants to add bending potentials. What do you think?

testsuite/hydrogel_builder_with_angles.py
```diff
@@ -0,0 +1,186 @@
+#
```
pm-blanco, line 1: I think that this test should check the angle instance counts:

1) For the case where no bending is added in the nodes:
Angle instance counts:
C-C-C    32
2) For the case when bending is added in the nodes:
Angle instance counts:
C-C-C    32
C-C-N    32
C-N-C    48

samples/build_angle.py
```diff
@@ -0,0 +1,118 @@
...
+# Example: demonstrating angle potential support in pyMBE.
+#
+# This script shows two complementary usages of `gen_angle=True`:
+#   1. create_residue with gen_angle=True
+#        - builds a single residue with auto-generated angles
+#   2. create_molecule with gen_angle=True
+#        - builds a chain of residues; angles are auto-generated across the
+#          full molecule, including angles that span residue boundaries
+#
+# Note: angle auto-generation is currently supported only in `create_residue`
```
pm-blanco: Remove this note, now it is actually supported :)

samples/build_angle.py
```diff
@@ -0,0 +1,118 @@
+#
```
pm-blanco: Have you tested if the bending potentials are correctly added in the case of molecules with nexted residues, i.e. when you have a reside as a side chain?

testsuite/angle_tests.py
```diff
@@ -0,0 +1,790 @@
+#
+# Copyright (C) 2024-2026 pyMBE-dev team
```
pm-blanco: Copyright here should only display the current year

pm-blanco (changes requested):
I have a few final formal revisions to the PR, but overall it looks really good. Thank you very much @1234somesh for an excellent work, I really like how the implementation looks.

samples/build_angle.py
1234somesh: Yes, an example including nested residue is added in samples/build_angle.py

samples/build_hydrogel_angle.py
1234somesh: build_hydrogel_angle.py is removed and merged with build_hydrogel.py. Now build_hydrogel.py serves dual purpose:  original hydrogel builder for lattice visualiztion and the new hydrogel builder which support angle generation

samples/build_hydrogel_angle.py
1234somesh: This is subtle point, which needs to be explicit I think. The alphabetic sorting is applied on the side particles only. So the angle with central particle "D", will remain C-D-C and not C-C-D after sorting. I have used three different particles "A","C","D" in the sample script samples/build_hydrogel.py and users can use the option "--inlcude" to inlude crosslinker angle generation when this script is exected and see this angle key ordering when the dataframe is printed.

samples/build_angle.py
```diff
@@ -1,17 +1,32 @@
 #
+# Copyright (C) 2026 pyMBE-dev team
```
pm-blanco, line 2: perhaps we could rename this sample to something more explicit like `setup_bending_potential.py`

samples/build_angle.py
```diff
@@ -1,17 +1,32 @@
...
 #        - builds a single residue with auto-generated angles
 #   2. create_molecule with gen_angle=True
 #        - builds a chain of residues; angles are auto-generated across the
 #          full molecule, including angles that span residue boundaries
-#
-# Note: angle auto-generation is currently supported only in `create_residue`
-# and `create_molecule`. Other higher-level builders (e.g. `create_hydrogel`)
-# do not yet propagate `gen_angle`.
-#
+#   3. create_residue with a nested residue as side chain
```
pm-blanco: Would not make more logical sense to move Demo 3 to Demo 2? So we move up in the pyMBE hierarchy of molecules:
Example 1: Simple residue
Example2: Residue with a nested residue
Example3: A molecule

samples/build_hydrogel.py
```diff
@@ -15,94 +15,241 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+# Demonstrates:
+#   1. Original hydrogel build (mpc=40, two bead types, no angles) with a 3D plot.
```
pm-blanco: Maybe replace "original" for "simple" hydrogel build (in the future no one will remember what "original" meant").  Instead of "no angles" I would add "no angular potential".

samples/build_hydrogel.py
```diff
@@ -15,94 +15,241 @@
...
-fig = plt.figure()
-ax = fig.add_subplot(111,projection="3d")
-lattice_builder.draw_lattice(ax=ax, 
-                             pmb=pmb)
-lattice_builder.draw_simulation_box(ax)
-plt.legend(fontsize=12)
-plt.show()
+
+def run_original_hydrogel(espresso_system):
+    """Build and visualize the original hydrogel (mpc=40, no angles)."""
```
pm-blanco: Refactor this note to match the above. Document all arguments of the function.

samples/build_hydrogel.py
```diff
@@ -15,94 +15,241 @@
...
+
+    fig = plt.figure()
+    ax = fig.add_subplot(111, projection="3d")
+    lattice_builder.draw_lattice(ax=ax, pmb=pmb)
+    lattice_builder.draw_simulation_box(ax)
+    plt.legend(fontsize=12)
+    plt.show()
+
+
+def define_angle_hydrogel_system(espresso_system, include_crosslinker_angles):
```
pm-blanco: Document all arguments of the function.

samples/build_hydrogel.py
```diff
@@ -15,94 +15,241 @@
...
+    pmb.define_hydrogel(name="simple_hydrogel",
+                        node_map=node_topology,
+                        chain_map=chain_topology)
+    hydrogel_id = pmb.create_hydrogel(name="simple_hydrogel",
+                                      espresso_system=espresso_system,
+                                      gen_angle=True)
+    return pmb, hydrogel_id
+
+
+def summarize_angles(pmb):
```
pm-blanco, line 157: Document all arguments of the function.

samples/build_hydrogel.py
```diff
@@ -15,94 +15,241 @@
...
+    """Return angle templates, instances, and counts by canonical angle name."""
+    angle_templates_df = pmb.get_templates_df(pmb_type="angle")
+    angle_instances_df = pmb.get_instances_df(pmb_type="angle")
+    angle_counts = angle_instances_df["name"].value_counts().sort_index()
+    return angle_templates_df, angle_instances_df, angle_counts
+
+
+def run_angle_case(label, include_crosslinker_angles, expected_angle_names, espresso_system):
+    """Clear the system, build one angle-hydrogel case, and assert expected angle names."""
```
pm-blanco: Document all arguments of the function.

samples/build_hydrogel.py
pm-blanco: For clarity, consider renaming to `define_hydrogel_with_angular_potential`

samples/build_hydrogel.py
```diff
@@ -15,94 +15,241 @@
...
+
+def summarize_angles(pmb):
+    """Return angle templates, instances, and counts by canonical angle name."""
+    angle_templates_df = pmb.get_templates_df(pmb_type="angle")
+    angle_instances_df = pmb.get_instances_df(pmb_type="angle")
+    angle_counts = angle_instances_df["name"].value_counts().sort_index()
+    return angle_templates_df, angle_instances_df, angle_counts
+
+
+def run_angle_case(label, include_crosslinker_angles, expected_angle_names, espresso_system):
```
pm-blanco: Consider renaming to `run_hydrogel_with_bending_potential`

samples/build_hydrogel.py
```diff
@@ -15,94 +15,241 @@
...
-
-fig = plt.figure()
-ax = fig.add_subplot(111,projection="3d")
-lattice_builder.draw_lattice(ax=ax, 
-                             pmb=pmb)
-lattice_builder.draw_simulation_box(ax)
-plt.legend(fontsize=12)
-plt.show()
+
+def run_original_hydrogel(espresso_system):
```
pm-blanco: consider renaming to `run_simple_hydrogel`

pyMBE/pyMBE.py
```diff
@@ -1452,7 +1574,246 @@ def define_default_bond(self, bond_type, bond_parameters):
                                bond_type=bond_type)
         tpl.name = "default"
         self.db._register_template(tpl)
-    
+
+    def define_angle(self, angle_type, angle_parameters, particle_triplets):
```
pm-blanco: I think it would be more clear to rename this function to `define_angular_potential` than simply `define_angle`, although I see the point in the analogy with `define_bond`. To me to "define an angle" sounds a bit vague, but we could leave it with this name if you think is clear enough.

pyMBE/pyMBE.py
```diff
@@ -1452,7 +1574,246 @@ def define_default_bond(self, bond_type, bond_parameters):
...
+                                side_particle2=side2,
+                                parameters=parameters_tpl,
+                                angle_type=angle_type)
+            tpl._make_name()
+            if tpl.name in angle_names:
+                raise RuntimeError(f"Angle {tpl.name} has already been defined, please check the list of particle triplets")
+            angle_names.append(tpl.name)
+            self.db._register_template(tpl)
+
+    def define_default_angle(self, angle_type, angle_parameters):
```
pm-blanco: if we rename `define_angle` this one should be renamed accordingly.

pyMBE/pyMBE.py
```diff
@@ -1452,7 +1574,246 @@ def define_default_bond(self, bond_type, bond_parameters):
...
+            "phi_0": PintQuantity.from_quantity(q=angle_parameters["phi_0"],
+                                                expected_dimension="dimensionless",
+                                                ureg=self.units),
+        }
+        tpl = AngleTemplate(parameters=parameters_tpl,
+                            angle_type=angle_type)
+        tpl.name = "default"
+        self.db._register_template(tpl)
+
+    def create_angle(self, particle_id1, particle_id2, particle_id3, espresso_system, use_default_angle=False):
```
pm-blanco: and this one should then also be renamed accordingly.

pm-blanco (approved):
LGTM

PR #151, opened 2026-06-24
Increasing pyMBE interoperability (part I)

jorch28:
### **Summary**

Application of the strategy pattern to delegate the responsability of executing the simulation engine to a particular engine class (e.g `EspressoEngine`).

### **Added**

- `EspressoEngine` Class
- `EspressoSystemProtocol` Class to check that an the working engine corresponds to espresso.system instance
- reaction methods to `EspressoEngine`
- handy functions to `EspressoEngine`
- miscellaneous methods as (`calculate_net_charge`, `change_volume_and_re-scale_particles`,`enable_motion_of_rigid_object`,etc.)
-  method `add_instances_to_engine` to add particles, bonds and angles to Espresso

### **Modified**

- `create_particle`,`create_residue`, `create_molecule`,etc. to avoid an explicit dependence to Espresso
- unit_tests and functional_tests accordingly so that they introduce the particles, bonds and angles before choosing espresso as the working engine. Execute pmb methods from the simulation_engine when they depend on espresso.

### **Fixed**

In commit: https://github.com/pyMBE-dev/pyMBE/pull/151/commits/217d15a41a0fda0c05d7125646c31ae3132376cd    

-  class `DiamondLattice`. box_l=(self.mpc+1)*self.bond_l.magnitude / (np.sqrt(3)*0.25) 
-  method `_create_hydrogel_chain`. backbone vector=vec_between_nodes / (self.lattice_builder.mpc+1)


pyMBE/pyMBE.py
```diff
@@ -827,7 +781,7 @@ def create_added_salt(self, espresso_system, cation_name, anion_name, c_salt):
         if anion_charge >= 0:
             raise ValueError(f'ERROR anion charge must be negative, charge {anion_charge}')
         # Calculate the number of ions in the simulation box
-        volume=self.units.Quantity(espresso_system.volume(), 'reduced_length**3')
+        volume=self.units.Quantity(box_l[0]*box_l[1]*box_l[2], 'reduced_length**3') ### Changed espresso.volume() to box_l**3
```
jngrad:
```suggestion
        volume=self.units.Quantity(np.prod(box_l), 'reduced_length**3')
```
I find `np.prod()` more expressive and it avoids accidental typos in the vector indices

pyMBE/pyMBE.py
```diff
@@ -1139,11 +1097,13 @@ def create_molecule(self, name, number_of_molecules, espresso_system, list_of_fi
                     if list_of_first_residue_positions is None:
                         central_bead_pos = None
                     else:
+                        ### This seems like a bug 
+                        ### the variable central bead pos gets assigned the same value for the lengthh of list_of_first_residue_positions
                         for item in list_of_first_residue_positions:
                             central_bead_pos = [np.array(list_of_first_residue_positions[pos_index])]
```
jngrad, line 1104: I agree, this code looks strange. If the current behavior is intended, the loop is superfluous.

pyMBE/pyMBE.py
```diff
@@ -1723,24 +1683,23 @@ def get_angle_template(self, side_name1, central_name, side_name2, use_default_a
...
-        self.db.espresso_angle_instances[angle_template.name] = angle_inst
-        espresso_system.bonded_inter.add(angle_inst)
-        return angle_inst
+        if isinstance(self.simulation_engine,EspressoSimulation):
+            angle_inst=self.simulation_engine._get_angle_instance(angle_template=angle_template)
+            return angle_inst
+        elif isinstance(self.simulation_engine,LammpsSimulation):
+            raise NotImplementedError('It has not yet been implemented for Lammps')
+        else:
+            raise RuntimeError('You have not set up any simulation engine yet')
```
jngrad, line 1702: Here and elsewhere, what is the intention behind the if/else chain? If the simulation engines share the same API, we could call `self.simulation_engine.method()` and let the class instance raise a `NotImplementedError` if the method doesn't exist for that engine, or a `RuntimeError` if the engine was never set, e.g. using a dummy engine class that throws for any method call. This would satisfy the [open–closed principle](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle): adding a new GROMACS simulation engine would only require implementing the corresponding class, and there would be no need to adapt all if/else chains in methods of `pyMBE.py`.

tutorials/pyMBE_tutorial.ipynb
```diff
@@ -3474,87 +6591,48 @@
...
...[39m \u001b[43m                            \u001b[49m\u001b[43mnumber_of_molecules\u001b[49m\u001b[43m=\u001b[49m\u001b[43m  \u001b[49m\u001b[43mN_peptide\u001b[49m\u001b[43m,\u001b[49m\n\u001b[32m      3\u001b[39m \u001b[43m                            \u001b[49m\u001b[43mespresso_system\u001b[49m\u001b[43m \u001b[49m\u001b[43m=\u001b[49m\u001b[43m \u001b[49m\u001b[43mespresso_system\u001b[49m\u001b[43m,\u001b[49m\n\u001b[32m      4\u001b[39m \u001b[43m                            \u001b[49m\u001b[43mlist_of_first_residue_positions\u001b[49m\u001b[43m \u001b[49m\u001b[43m=\u001b[49m\u001b[43m \u001b[49m\u001b[43m[\u001b[49m\u001b[43m[\u001b[49m\u001b[43mBox_L\u001b[49m\u001b[43m.\u001b[49m\u001b[43mto\u001b[49m\u001b[43m(\u001b[49m\u001b[33;43m'\u001b[39;49m\u001b[33;43mreduced_length\u001b[39;49m\u001b[33;43m'\u001b[39;49m\u001b[43m)\u001b[49m\u001b[43m.\u001b[49m\u001b[43mmagnitude\u001b[49m\u001b[43m/\u001b[49m\u001b[32;43m2\u001b[39;49m\u001b[43m]\u001b[49m\u001b[43m*\u001b[49m\u001b[32;43m3\u001b[39;49m\u001b[43m]\u001b[49m\u001b[43m)\u001b[49m\n\u001b[32m      5\u001b[39m pmb.get_instances_df(pmb_type=\u001b[33m\"\u001b[39m\u001b[33mpeptide\u001b[39m\u001b[33m\"\u001b[39m)\n",
```
jngrad, line 3542: Here and elsewhere in this notebook: that's a lot of text formatted for specific interpreters ([ANSI escape code](https://en.wikipedia.org/wiki/ANSI_escape_code)). Do we really need to ship this notebook with all outputs? There are ways to execute notebooks in CI to regenerate output cells, and to export the final notebook as a HTML page for the GitHub Pages.

pyMBE/simulation_builder/base_engine.py
```diff
@@ -0,0 +1,144 @@
+from abc import ABC,abstractmethod
```
jngrad, line 19: Here and in other new files: missing copyright notices.

pyMBE/simulation_builder/espresso_engine.py
```diff
@@ -0,0 +1,1453 @@
...
+        
+        Returns:
+            espresso_particles_id(list): list of pids of the particles that are saved in espresso
+        """
+        espresso_particles=self.espresso_system.part.all()
+        return espresso_particles.id    
+    
+    
+    def _get_particle_pos_espresso(self,id):
+        """_summary_
```
jngrad: Here and elsewhere, there are numerous occurrences of `_summary_`. Are these todo notes?

pyMBE/exceptions/pmb_warnings.py
```diff
@@ -0,0 +1,17 @@
+import warnings
+from functools import wraps
+
+def deprecated(new_function):
+    """Wrapper function to display a warning deprecation message to all functions that are going to be removed in a future
+    """
```
jngrad, line 24: I'm a bit uneasy about pyMBE re-implementing a feature already provided by `@warnings.deprecated()`. I understand that it's only available in Python 3.13 onward, but Python 3.12 is old. It's no longer receiving bugfixes, only security fixes, and in about 9 months it won't be supported by new packages in the NumPy ecosystem ([NEP 29 drop table](https://numpy.org/neps/nep-0029-deprecation_policy.html#drop-schedule)). Ubuntu 24.04 LTS is the last OS that still imposes Python 3.12, and a couple of HPC clusters are stuck with Python 3.11 and 3.12.

We should discuss in the next meeting to which degree this is an issue for our users. Ubuntu 26.04 LTS is already available and the ICP plans to migrate to it in August. For HPC clusters where EESSI is available, Python 3.13 can be obtained from the 2025 stacks. On HPC clusters that don't provide EESSI, I build Python 3.14 from sources since older Python versions can be tricky to work with in conjunction with Cython and NumPy. Users of conda-like environments can install a new Python version with an extra command line option. I won't oppose this re-implementation, it's not a dealbreaker for me, but I think it's worth considering the maintenance aspects of re-implementing a functionality that is now widely available.

jngrad (commented):
Overall, I really like how the ESPResSo instance was made a data member of the pyMBE class. This really helps separate the pyMBE molecular building logic from the ESPResSo implementation details, and will make it easier for contributors from other communities to contribute to pyMBE, once LAMMPS is supported.

pyMBE/pyMBE.py
jorch28: Seems like a great idea to me!  
To proceed with the modification:
1) By default the instance variable of the simulation_engine=DummySimulationengine() which overrides the methods from the BaseClass and raises a RuntimeError in case that the method is executed?
2) I will write the methods from the LammpsSimulationEngine with a statement that raises a NotImplementedError, so the implementation follows the open-closed principle

jngrad: Regarding point 1, here is the general pattern:
```py
class MyClass:
    def test(self):
        return 5
    def __getattr__(self, attr):
        if attr not in self.__dict__:
            raise RuntimeError('You have not set up any simulation engine yet')
        return super().__getattr__(attr)


obj = MyClass()
obj.test() # doesn't raise
obj.method() # raises RuntimeError
```
I'm not sure how this interact with ABC metaclass options.

pyMBE/exceptions/pmb_warnings.py
jorch28: I implemented this function so users now are warned that they have to move their usage of the function `setup_langevin_dynamics`, `relax_espresso_system`,`setup_electrostatic_interactions` , etc. To be employed by the simulation_engine as `simulation_engine.setup_langevin_dynamics`, `simulation_engine.relax_espresso_system`,etc. In order that when they start using the new release version not all of their old code breaks due to these changes being implemented.

Respect to upgrading pyMBE to python 3.13 we may find some troubles related to incompatibilities between the versions of the external python libraries, although maybe this is not a major issue, as the differences are between 3 minor versions .

pyMBE/simulation_builder/espresso_engine.py
jorch28: Nope. The extension AutoDocstring from VSCode generates the word __summary__  automatically. I  almost always keep it, because it is the header of the section as it happens as well with the words: Args, Return and sometimes Notes.

tutorials/pyMBE_tutorial.ipynb
jorch28: No, this has been done unintentionally. Now I'm going to delete the outputs.

pyMBE/pyMBE.py
```diff
@@ -885,32 +836,31 @@ def create_bond(self, particle_id1, particle_id2, espresso_system, use_default_b
         bond_tpl = self.get_bond_template(particle_name1=particle_inst_1.name,
                                           particle_name2=particle_inst_2.name,
                                           use_default_bond=use_default_bond)
-        bond_inst = self._get_espresso_bond_instance(bond_template=bond_tpl,
-                                                    espresso_system=espresso_system)
-        espresso_system.part.by_id(particle_id1).add_bond((bond_inst, particle_id2))
+        # bond_inst = self._get_espresso_bond_instance(bond_template=bond_tpl,
```
pm-blanco, line 840: This comments should be removed

pyMBE/pyMBE.py
pm-blanco: I agree too, the loop should be removed and central_bead_pos should simply take the first value in `list_of_first_residue_positions`

pyMBE/pyMBE.py
```diff
@@ -1686,10 +1646,10 @@ def create_angular_potential(self, particle_id1, particle_id2, particle_id3, esp
                                             central_name=particle_inst_2.name,
                                             side_name2=particle_inst_3.name,
                                             use_default_angle=use_default_angle)
-        angle_inst = self._get_espresso_angle_instance(angle_template=angle_tpl, espresso_system=espresso_system)
+        # angle_inst = self._get_espresso_angle_instance(angle_template=angle_tpl, espresso_system=espresso_system)
```
pm-blanco, line 1649: remove this comment

pyMBE/pyMBE.py
```diff
@@ -1686,10 +1646,10 @@ def create_angular_potential(self, particle_id1, particle_id2, particle_id3, esp
...
 
         # ESPResSo angle bonds are added to the central particle
-        espresso_system.part.by_id(particle_id2).add_bond((angle_inst, particle_id1, particle_id3))
+        # espresso_system.part.by_id(particle_id2).add_bond((angle_inst, particle_id1, particle_id3))
```
pm-blanco, line 1652: remove this comment

samples/Landsgesell2022/time_series/analyzed_data.csv
pm-blanco, line 1: time series should not be comitted to the main repository please remove this file and other similar ones

pyMBE/simulation_builder/espresso_engine.py
pm-blanco: I think that the place holder "_summary" in the docstrings should be replaced by a short description of what each method does.

pyMBE/exceptions/pmb_warnings.py
jngrad: pyMBE meeting discussion: keep this re-implementation for now, since it will be a couple of months until our community fully migrates to the newer Python version, but replace it with `@warnings.deprecated()` from Python 3.13 before the next minor release of pyMBE.

jorch28:
@pm-blanco @jngrad 
For version **ESPREesSo 5.0-dev** testsuite is failing due to an error importing `espressomd`. In the workflow execution logs it is shown that the module espresso is not loaded. I think this happens because in the **push_pull.yaml** the value for the key `upload artifact` in  `jobs.ubuntu.strategy.matrix.espresso.upload_artifact`=False.

For **ESPREsSO 4.2.2**, all tests are passing, but they last longer than 300s and the last test 26/26 is not being executed.  Did this happened before?

jngrad:
There seems to be an issue with how `module save` handles the `dev.eessi.io` stack. In addition, the runner is executing actions with the wrong Node.js version. Let's drop the development stack in favor of the stable repository, and upgrade all software and operating systems. I'll open a separate PR, since it's orthogonal to this work.

pm-blanco:
@jorch28 we have already updated the runner pipeline in the main pyMBE-dev repo, can you merge it here and see if now the tests pass?

jorch28:
Now, for version 5.0.1 seems that there is no error related to  import espressomd anymore. Thank you for your very fast answer! I'm going to add the method _check_bond_input to the LammpsSimulationEngine that is missing  and therefore is not complying with the contract setup by the interface.
The **actual problem** was related to the differences between the methods and attributes from `espressosystem` in **version 4.2.2** and  **version 5.0.1**

pyMBE/pyMBE.py
jorch28: Then, do I modify it in this current pull request as Pablo is suggesting? Or do i set an Issue?

pm-blanco: @jorch28 please just modify it in this pull request.

jngrad:
Here is a blueprint for testing the LAMMPS engine in CI: jngrad/pyMBE@40ad03ca61. One could adapt existing test cases to run with both ESPResSo and LAMMPS (in this case, move the `@ut.skipIf(not is_engine_available` from above the class to above the methods), or create separate test files for each engine (I wouldn't recommend: that could lead to diverging code).

LAMMPS isn't available in foss/2025a yet.

pm-blanco:
@jngrad @jorch28 I addressed all of our review comments and cleaned up a number of other pending issues I found while addressing the others. I think that now it should be close or ready to be merged, but please take a look so we can merge it ASAP. I removed the duplicated functions with the custom deprecation warnings because I do not think that makes sense to maintain an obsolete API when we are close to deploy a new release to pyMBE in any case.

jngrad (approved):

PR #152, opened 2026-07-02
Upgrade runner environment

jngrad:


jngrad:
I gave the matrix options more generic names, so that we can later decide to write `eessi_modules: ESPResSo/4.2.2-foss-2023b LAMMPS/29Aug2024-foss-2023b-kokkos` to get both ESPResSo and LAMMPS in the same testsuite, or any other package that is too time-consuming to install via `pip install`. Alternatively, we could modify the job matrix with LAMMPS-only packages; that would be cleaner in my opinion, since pyMBE should be able to run with either engine even when the other engine is missing.

testsuite/CTestTestfile.cmake
```diff
@@ -18,7 +18,7 @@
 #
 
 cmake_minimum_required(VERSION 3.22.1)
-find_package(Python 3.8 REQUIRED COMPONENTS Interpreter NumPy)
+find_package(Python 3.8 REQUIRED COMPONENTS Interpreter)
```
jngrad, line 21: The NumPy option was actually superfluous. It instructs CMake to search for NumPy header files, but those are not actually needed, since we are not compiling any source code. The presence of `NumPy` here can trigger a CMake runtime error on HPC clusters under unclear (and hard to reproduce) circumstances.

pm-blanco (approved):

pm-blanco:
LGTM

PR #154, opened 2026-08-14
fix functional tests

pm-blanco:
Fixes issues with the functional tests `cph_ideal_tests` and `weak_polyelectrolyte_dialysis_test`:

1) `cph_ideal_tests`  was failing due to a small discrepancy in one data point  (0.19952377 absolute error) 
- It was fixed by simply slightly increasing the sampling for the test. This has no significant impact on the overall run time of the functional costs since this test is still the less expensive.
- A curious observation is that this test ran without issue on Espresso 4.2.2 but started failing when we migrated to 5.0.0. Might simply reflect a change in the rng as @jngrad suggested. 

2) `weak_polyelectrolyte_dialysis_test` was failing due to a regresion introduced in #151 when a reference data set was removed by mistake.


jngrad:
samples CI passes: https://github.com/jngrad/pyMBE/actions/runs/31796016858/job/94753193676

jngrad (approved):
LGTM

PR #155, opened 2026-08-26
Drop support for ESPResSo 4.2

jngrad:
Remove compatibility functions for ESPResSo 4.2. Add support for ESPResSo 5.1-dev.


jngrad:
Here is the samples testsuite: https://github.com/jngrad/pyMBE/actions/runs/32992804738

pm-blanco (approved):
LGTM, one question @jngrad is pydantic>2.0.0 supported by espressomd>5.0.0? If yes we should try alleviate our dependency on the legacy version of pydantic

jngrad:
AFAIK ESPResSo doesn't rely on pydantic, so an update shouldn't hurt.
I just tried v2 and it looks like we would have to update the pyMBE storage:
```
  File "/work/jgrad/pyMBE/pyMBE/storage/instances/bond.py", line 24, in <module>
    class BondInstance(PMBBaseModel):
pydantic.errors.PydanticUserError: The `field` and `config` parameters are not available in Pydantic V2,
please use the `info` parameter instead.
```

pm-blanco:
As far as I remember, the problem we had is that Pydantic 2.0+ had some python requirements that were not compatible with those of espresso 4.2. Indeed, we would need to adapt a few legacy pydantic parameters that I had to introduce in the pyMBE storage but I think would not be too dificult. In any case, this goes beyond this PR, I just wanted to know if there was any known caveat on upgrading Pydantic 2.0+ on the espresso side :)

jngrad:
We could bump all Python requirements to those available in Ubuntu 25.10 and later, but then we would have to drop support for NumPy 1.26, which is still available as the default NumPy version on Ubuntu 24.04 LTS.

```diff
diff --git a/README.md b/README.md
index e4bac0f..4c34583 100644
--- a/README.md
+++ b/README.md
@@ -72,3 +72,3 @@ pyMBE supports ESPResSo 5.0 and ESPResSo 5.1-dev.
 Following the NEP29 guidelines, we recommend using Python3.11+.
-Both NumPy 1 and NumPy 2 are supported.
+Only NumPy 2 is supported.
 
@@ -86,3 +86,3 @@ python3 -m venv pymbe  # create a local folder named pymbe containing the enviro
 source pymbe/bin/activate  # activate the virtual environment
-python3 -m pip install -r requirements.txt "numpy<2.0" "pandas<2.0"
+python3 -m pip install -r requirements.txt
 python3 maintainer/configure_venv.py --espresso_path=/home/user/espresso/build # please adapt the espresso path accordingly
@@ -92,8 +92,2 @@ deactivate  # deactivate the virtual environment
 
-NumPy 2 users should adapt the pip command as follows:
-
-```sh
-python3 -m pip install -r requirements.txt "numpy>=2.1" "pandas>=2.0"
-```
-
 We highlight that the path `/home/user/espresso/build` is just an example of a possible path to the ESPResSo build folder. 
diff --git a/requirements.txt b/requirements.txt
index 427f169..85f345d 100644
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,11 +1,11 @@
-numpy>=1.23
-pandas>=1.5.3
-pint>=0.20.01
-pint-pandas>=0.3
-biopandas==0.5.1.dev0
-scipy>=1.8.0
-matplotlib>=3.5.1
-pydantic<2.0.0
+numpy>=2.2.4
+pandas>=3.0
+pint>=0.25.2
+pint-pandas>=0.8.0
+biopandas==0.5.2
+scipy>=1.15.3
+matplotlib>=3.6.3
+pydantic>=2.10.6
 # soft dependencies to run the samples
-tqdm>=4.57.0
+tqdm>=4.66.2
 # soft dependencies to run the testsuite
```

pm-blanco:
@jngrad let's discuss it on the next pyMBE meeting and see if our community would be okay with that.

